Issue #186, opened 2018-05-24
Create Template Modal


velkymx:
Need to implement a modal for templates. Include a WYSIWYG editor.

Web link: https://zpl.io/aMm7N7z

<img width="498" alt="screen shot 2018-05-31 at 5 10 39 pm" src="https://user-images.githubusercontent.com/39599643/40814539-a1591e20-64f5-11e8-80ec-d4b66f59c8cb.png">

milaendo:
https://github.com/ProcessMaker/bpm/pull/246

PR #246, opened 2018-06-07
#186 template modal created/ using tinymce and cancel button function…

milaendo:
under templates modal is created


milaendo:
screen shot 2018-06-07 at 11 59 09 am

PR #270, opened 2018-06-10
Resolve issue with dusk browser test. Refactor task listing widget to…

tdondich:
… have process uid passed in as a prop instead of parsing from url.


tdondich:
Merged without hesitation.  ;)  Had to get develop branch working again.

Issue #230, opened 2018-06-06
Refresh Login


velkymx:
We need to refresh the login page to more closely match the design. Must use Bootstrap layout.

![screenshot 2018-06-06 13 22 00](https://user-images.githubusercontent.com/535300/41063112-9ebbc88e-698c-11e8-9c3b-50f139ed16ed.png)

Full spec with layout details:
https://app.zeplin.io/project/5a872c6ffc8c7ad639253eb9/screen/5b0e6a698ddbd152620f9ec7

velkymx:
Merged

PR #271, opened 2018-06-11
added missing links in login page

uditguru19:
First Commit - Added missing links. References to #230

### DIFF
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -2,80 +2,96 @@
 
 @section('content')
 <div class="container mt-5">
-  <div class="row">
+    <div class="row">
 
-    <div class="col">
-    </div>
+        <div class="col">
+        </div>
 
-    <div class="col-6">
-      <div align="center">
-        <img class="p-5" src="/img/processmaker-login-logo.png">
-      </div>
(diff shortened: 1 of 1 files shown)


resources/views/auth/login.blade.php
```diff
@@ -2,80 +2,96 @@
...
+
+                <a class="btn-link text-primary pl-0" href="{{ route('password.request') }}">
+                    {{ __('Forgot Password?') }}
+                </a>
+                <div class="center mt-5">
+                    <p class="text-center mb-0">Dont have an account?</p>
+                </div>
+                <div class="text-center">
+                    <a class="btn-link text-primary pl-0" href="">
+                        Start your free trial now!
```
velkymx: This needs to use the translation class like the "Forgot Password"

resources/views/auth/login.blade.php
```diff
@@ -2,80 +2,96 @@
...
-          {{ __('Forgot Password?') }}
-        </a>
-      </form>
-    </div>
 
-    <div class="col">
     </div>
+    @endsection
+
+    @section('css')
```
velkymx: Remove this section. Not needed if there isn't any custom CSS

PR #295, opened 2018-06-14
Feature/291

milaendo:
screen shot 2018-06-14 at 10 38 07 am


milaendo:
this ticket has a temporary path at /test. Taylor has suggested that there be a second ticket for creating the controller and route that would actually be used for this page.

velkymx:
Closing PR for this until we get the backend functionality added.

PR #314, opened 2018-06-19
Disable Dusk in Prod

### LINKED ISSUE #267: Prevent DuskServiceProvider from loading when APP_ENV is set to production
DuskServiceProvider should not be auto-found but instead, loaded on demand when APP_ENV is not production.

velkymx:
Updated autoloader and service provider registration to fix #267

### DIFF
--- a/ProcessMaker/Providers/ProcessMakerServiceProvider.php
+++ b/ProcessMaker/Providers/ProcessMakerServiceProvider.php
@@ -45,7 +45,7 @@ class ProcessMakerServiceProvider extends ServiceProvider
     {
 
         // Dusk, if env is appropriate
-        if ($this->app->environment('local', 'testing', 'development')) {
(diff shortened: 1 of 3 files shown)


ProcessMaker/Providers/ProcessMakerServiceProvider.php
```diff
@@ -43,6 +43,12 @@ public function boot()
      */
     public function register()
     {
+
+        // Dusk, if env is appropriate
+        if ($this->app->environment('local', 'testing', 'development')) {
```
tdondich: Incorrect usage.  You need to pass in the diff environments with an array.  Please see: https://laravel.com/docs/5.6/configuration#determining-the-current-environment

Also, might be good to simply do:

if(!$this->app->environment('production')) {
...
}

PR #315, opened 2018-06-19
Feature/307

marcoAntonioNina:
Add api User profile

api/1.0/admin/profile GET/PUT

issue https://github.com/ProcessMaker/bpm/issues/307

### DIFF
--- a/ProcessMaker/Http/Controllers/Api/Administration/UsersController.php
+++ b/ProcessMaker/Http/Controllers/Api/Administration/UsersController.php
@@ -18,7 +18,7 @@ class UsersController extends Controller
 
     /**
      * Fetch a collection of users based on paged request and filter if provided
-     * 
+     *
      * @return JsonResponse A list of matched users and paging data
      */
     public function index(Request $request)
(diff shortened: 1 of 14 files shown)


routes/api.php
```diff
@@ -31,6 +31,10 @@
             Router::get('users', 'Administration\UsersController@index');
             Router::get('users/{user}', 'Administration\UsersController@get');
 
+            //User profile
+            Router::get('admin/{user}/profile', 'Administration\UsersController@profile');
+            Router::put('admin/{user}/profile', 'Administration\UsersController@update');
+
```
velkymx: The logged in use should be able to directly poll the api end point without an id. 

ie: api/1.0/admin/profile

These other routes should be the same. This is different from the admin/user end points.

marcoAntonioNina: Changed

ProcessMaker/Http/Controllers/Api/Administration/UsersController.php
```diff
@@ -2,8 +2,13 @@
```
velkymx: I'd like to move this controller to User/ProfileController there will be an Admin/UserController for the admin area.

ProcessMaker/Http/Controllers/Api/Administration/UsersController.php
```diff
@@ -74,4 +79,82 @@ public function get(Request $request, User $user)
...
+    /**
+     * Load profile user
+     *
+     * @return JsonResponse
+     */
+    public function profile()
+    {
+        $url = '';
+        $user = User::find(Auth::id());
+        if (!empty($user->avatar) && Storage::disk('profile')->exists($user->avatar)) {
```
velkymx: How do we retrieve the user's avatar image? Is there an image url?

Issue #334, opened 2018-06-22
Forgot Password 


meadrat:
Re: https://github.com/ProcessMaker/bpm/issues/291 - 2 changes needed to Forgot Password screen

<img width="404" alt="screen shot 2018-06-22 at 2 01 13 pm" src="https://user-images.githubusercontent.com/39599643/41799039-c43a9500-7624-11e8-8f34-8af6aa6b8531.png">

1. Add a '?' so it reads 'Forgot Your Password?'

<img width="221" alt="screen shot 2018-06-22 at 1 57 19 pm" src="https://user-images.githubusercontent.com/39599643/41798962-76bdfeb6-7624-11e8-96f6-2a185a986635.png">

2. Delete the dash in Email so it reads 'Email Address'

<img width="121" alt="screen shot 2018-06-22 at 1 57 47 pm" src="https://user-images.githubusercontent.com/39599643/41798961-76a22998-7624-11e8-821b-4c5e066112bb.png">

PR #335, opened 2018-06-25
Login, Reset and Password Template Updates

velkymx:
There are 4 templates overhauled to match the design provided by @meadrat in issue #334 

There are routes for all three pages.

/login
/password/reset
/password/newpass
/password/success

Design Samples: 

Login: https://zpl.io/VQqAepW
Password Reset: https://zpl.io/a7ykoYW
New pass: https://zpl.io/aN0oe7n
Password Success: https://zpl.io/awY18p7

NOTE: I made the reset, success and new pass forms a little wider and consistent.

### DIFF
--- a/resources/views/auth/passwords/reset.blade.php
+++ b/resources/views/auth/passwords/reset.blade.php
@@ -5,7 +5,7 @@
   <div class="formContainer">
     <img src="/img/md-blue-logo.png">
     <h3>{{__('Reset Your Password')}}</h3>
-    <form role="form" method="POST" action="{{ url('/password/reset') }}">
+    <form role="form" class="form" method="POST" action="{{ url('/password/reset') }}">
       {{ csrf_field() }} {{-- Needs to be enable when we hook up Controllers <input type="hidden" name="token" value="{{ $token }}"> --}}
       <div class="form-group{{ $errors->has('password') ? ' has-error' : '' }}">
         <label for="password">{{__('New Password')}}</label>
@@ -29,8 +29,9 @@
     width: 400px;
   }
 
-  .formContainer form {
+  .formContainer .form {
     margin-top: 50px;
+    text-align: left;
   }
 
   .formContainer h3 {
--- a/resources/views/auth/passwords/reset.blade.php
+++ b/resources/views/auth/passwords/reset.blade.php
@@ -4,7 +4,7 @@
 <div align="center">
   <div class="formContainer">
     <img src="/img/md-blue-logo.png">
-    <h3>{{__('Forgot Your Password?')}}</h3>
+    <h3>{{__('Reset Your Password')}}</h3>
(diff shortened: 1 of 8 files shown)


resources/views/auth/passwords/reset.blade.php
```diff
@@ -1,70 +1,72 @@
...
+
+@section('css')
+<style lang='scss' scoped>
+  img{
+    margin-top: 90px;
+  }
+  form{
+    width: 304px;
+  }
+  .header1{
```
tdondich: Not an ideal class name.  Please revise.

resources/views/auth/passwords/reset.blade.php
```diff
@@ -1,70 +1,72 @@
...
+    font-size: 18px;
+    font-weight: 500;
+  }
+  .header2{
```
tdondich: Not a good class name, please revise.

resources/views/auth/passwords/reset.blade.php
```diff
@@ -1,70 +1,72 @@
...
+<style lang='scss' scoped>
+  img{
+    margin-top: 90px;
+  }
+  form{
+    width: 304px;
+  }
+  .header1{
+    font-size: 18px;
+    font-weight: 500;
```
tdondich: We have a font-weight of 500 available?  First time I've seen it.

resources/views/auth/passwords/reset.blade.php
```diff
@@ -1,70 +1,72 @@
...
+      </div>
+    </div>
+    <div class="col">
     </div>
+  </div>
 </div>
 @endsection
+
+@section('css')
+<style lang='scss' scoped>
```
tdondich: This is in a blade template, not a vue component.  Scoped is not available here, lang is not allowed.

Ideally, this is put into a sep scss file called auth/reset.scss and then uses webpack to build.  Then this css section uses a <link> element to bring in that css.

Because it's not scoped, it's important that you prefix these items with an id that scopes these classes.  So something like #reset-password img, for example.

resources/views/auth/passwords/reset.blade.php
```diff
@@ -1,70 +1,72 @@
...
+                <input id="email" type="email" class="form-control{{ $errors->has('email') ? ' is-invalid' : '' }}" name="email" value="{{ $email or old('email') }}" required autofocus>
+                @if ($errors->has('email'))
+                    <span class="invalid-feedback">
+                        <strong>{{ $errors->first('email') }}</strong>
+                    </span>
+                @endif
             </div>
-        </div>
+            <button type="submit" class="btn btn-secondary btn-block mt-3 mb-3 text-light text-uppercase">
+              {{ __('REQUEST RESET LINK') }}
```
tdondich: Don't use all caps.  If it needs to render with all caps, uses CSS text-transform rules.

Issue #288, opened 2018-06-13
Requests Menu + Route


velkymx:
The requests menu needs to be configured as below. We also need a route defined for the main menu "Requests" (ie http://bpm4.processmaker.local/requests)

![screenshot 2018-06-13 09 41 50](https://user-images.githubusercontent.com/535300/41365341-02a12658-6eee-11e8-8319-064507a7fef6.png)

Design: https://app.zeplin.io/project/5aaa8e0656f3b51f3cf73864/screen/5ae368c32b04e703bf19fc94

velkymx:
https://processmaker.atlassian.net/browse/PMUI-134

meadrat:
These are the correct designs - Sergey needs to update his. 

https://zpl.io/29DMW1w 
https://zpl.io/aNpR84n

<img width="209" alt="screen shot 2018-06-14 at 6 33 58 pm" src="https://user-images.githubusercontent.com/39599643/41445989-e4c18ba4-7001-11e8-9a6e-3e77bf398fd2.png">
<img width="155" alt="screen shot 2018-06-14 at 6 33 49 pm" src="https://user-images.githubusercontent.com/39599643/41445991-e507ddd4-7001-11e8-8b3a-5e89a6b148a3.png">

meadrat:
@velkymx the designs you had above were outdated- I attached the correct ones.

tdondich:
And what does the drop down Do?  What if I have access to over 100
processes ?

milaendo:
https://github.com/ProcessMaker/bpm/pull/338

velkymx:
@meadrat what does the All Process button do? Is it a drop down? Do you have an example of it working?

meadrat:
This is what the expanded All Processes dropdown looks like @velkymx 

<img width="259" alt="screen shot 2018-06-28 at 9 43 43 am" src="https://user-images.githubusercontent.com/39599643/42048344-ddcb975c-7ab7-11e8-9428-7af4cbce2015.png">
<img width="257" alt="screen shot 2018-06-28 at 9 43 37 am" src="https://user-images.githubusercontent.com/39599643/42048345-dde3539c-7ab7-11e8-8516-6db5452d28b1.png">

PR #338, opened 2018-06-25
Feature/288

milaendo:
screen shot 2018-06-25 at 4 38 27 pm

I am unsure of the content inside of this page, it was there previously. The sidenav has been updated with the icomoon icons 

Ref #288

### DIFF
--- a/ProcessMaker/Http/Middleware/GenerateMenus.php
+++ b/ProcessMaker/Http/Middleware/GenerateMenus.php
@@ -231,28 +231,28 @@ class GenerateMenus
                 'label' => __('menus.sidebar_request.in_progress'),
                 'header' => false,
                 'route' => 'home',
-                'icon' => 'fa-user',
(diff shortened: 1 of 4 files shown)


routes/web.php
```diff
@@ -26,7 +26,7 @@
 
     $this->get('/request', function(){
```
velkymx: The route should be plural per the ticket

ProcessMaker/Http/Middleware/GenerateMenus.php
```diff
@@ -231,28 +231,28 @@ public function handle(Request $request, Closure $next)
                 'label' => __('menus.sidebar_request.in_progress'),
                 'header' => false,
                 'route' => 'home',
```
velkymx: Where is the All Processes entry?

milaendo:
screen shot 2018-06-27 at 9 23 40 am

Issue #377, opened 2018-07-11
Flow Improvement


velkymx:
Flow creation from Crown when drag
Enable Crown Delete
Anchor in flows
Embed shape in participant when drag from toolbar

PR #381, opened 2018-07-12
Process Flow Improvements #377

hjonathan:
Features:
- Flow creation from Crown when drag
- Enable Crown Delete
- Anchor in flows 
- Embed shape in participant when drag from toolbar

Ref #377


velkymx:
So good. One thing though...

Expectation: 

screenshot 2018-07-12 15 32 50

Reality:

screenshot 2018-07-12 15 32 54

Can we fix this?

Issue #370, opened 2018-07-06
Requests List


velkymx:
I want to click on the top navigation button and retrieve a list of Requests that are started. 

Link: https://zpl.io/bAvwMNp

![screenshot 2018-07-06 14 31 24](https://user-images.githubusercontent.com/535300/42401175-483677f2-8129-11e8-912d-33f1cd9fb90a.png)

This is the main In Progress requests page. When a user makes a Request, it will show up in this list. 

**Dashboard**

- There will be a dashboard at the top of all Tasks and Requests pages.
- Consists of OVERDUE, AT RISK and ON TIME + the amt of items that correspond to the filter
- The colors correspond to t-minus timing DUE DATES - they are buttons that are clickable and will filter the results
- If there are requests,
--- OVERDUE will be PM red - when the time is the exact due date/ time and after
--- AT RISK will be PM yellow - from 2 days/ 48 hours up until the exact due date/ time
--- ON TIME will be PM blue - from current time up to 2 days/ 48 hours before it is due
--- Empty state (if there are no requests) it will be gray w/ 0 requests

**Table Sections** - these are the default columns. User can change these with icon 
- ID (default sorted by ID) - the one being sorted by will be darker gray (#333333) with the arrow pointing down by default
- Task
---- insert light gray (#e9edf1) vertical line btw Task and Process
- Process
- Assigned to
--- Includes username with avatar photo
--- If no photo, will show user name’s first and last name initials
--- Light gray w/ generic avatar if Unassigned
- Due - shows t-minus timing as another way of showing when the task is due (i.e. 2 Days Ago, Tomorrow, Yesterday, 5 Days)
- Due Date - will have a colored dot on the left of the time that corresponds with its status - shows exact time (date + time)
- Created On - shows the exact time/ date that the task was created
... - shows Pause, Comment and Reassign mouseover - https://zpl.io/VqxAYDJ

**Search** as you type search capabilities

**Upper right-hand corner of table:**

- **Column icon** - user can click this icon to show all of the available columns that they can view. A pop-up will show with a list of columns and checkboxes to the left of each column title

- **Arrows** - these arrows will only show if the table is wider horizontally than the screen

meadrat:
@velkymx I added the link and requirements

caleeli:
@meadrat could you clarify what is "Due Date" and how it should be calculated. Does it depends on a process property or the tasks inside the process?

caleeli:
[...]

PR #386, opened 2018-07-17
Requests list, Solves #370

danloa:
- The list is located in the uri cases/requests
- There is a semaphorization bubble besides the due date
- A new api endpoint requests was added
- The search functionality of the requests is pending.


velkymx:
@caleeli I'd like you to refactor everything that references "Case" or "Cases" to use requests or request instead.

PR #387, opened 2018-07-18
CSS Clean Up for Vuetables

velkymx:
I've reworked the CSS for the vuetables to take advantage of the bootstrap variables. In doing so, I've also update the global font handling and updated the css to match.


resources/assets/sass/app.scss
```diff
@@ -144,7 +140,7 @@ body {
 }
 
 .form-group label {
-  font-size: 12px;
+  // font-size: 12px;
```
caleeli: remove this commented code

velkymx: Good call

Issue #369, opened 2018-07-06
Process List


velkymx:
When you click on the top navigation "Processes" I want to see a list of processes. Click through to the process designer.

Link: https://zpl.io/blm4xXX

<img width="1026" alt="screen shot 2018-07-12 at 3 11 15 pm" src="https://user-images.githubusercontent.com/39599643/42662268-e03b31b2-85e5-11e8-9260-c56762a2e3ac.png">

- **Search** as you type search capabilities - searches the process titles, category, status and created by

- **+ PROCESS** dark gray (#788793) button in the upper right-hand corner to add a Process. This button goes to https://zpl.io/VQYjPeR.

- **Columns** - these are the default columns but the user can determine which ones they want to see
--- Process - the title of the process, the user decides this when they + PROCESS, the maximum character length for a process title is x, if its longer than x characters, we will put ... and show the full title on scroll over
--- Category - choosing a category is mandatory for 4.0 (was optional in 3x)
--- Status - Active (green dot) or Inactive (red dot) or Draft (blue dot)
--- Modified By
--- Modified - date/ time
--- Created - date/ time
--- ... - Edit, Deactivate, Delete rollover icons - if user selects Edit, they can edit the title of the process

**Upper right-hand corner of table:**

- Column icon - user can click this icon to show all of the available columns that they can view. A pop-up will show with a list of columns and checkboxes to the left of each column title. 

- Arrows - these arrows will only show if the table is wider horizontally than the screen; the process column will always remain static (to the vertical line) and the others will move to the left and be hidden when the user clicks the arrow and the page moves to the right. 

**Pagination**
Pagination is determined by user. The user selects how many processes they would like to see per page and that setting is saved for when they return to the app. Default is 10 processes per page. User can select to view 10, 25, 50 or 100 processes per page. 

**Responsiveness** 
Need to determine responsive sizing requirements. I.e. if less than 800 pixels.. 

Not part of beta
- Filtering
- Process type
- Published status (yellow dot)

meadrat:
@velkymx I added requirements to this. Sergey needs to make some UI changes and then I will update the screenshot.

caleeli:
@velkymx could attach the screenshot of  https://zpl.io/VQYjPeR.
We didn't have access to zeplin

velkymx:
New process moved to own ticket #405

PR #396, opened 2018-07-19
Processes List, Solves #369

danloa:
- Added endpoint to return the xml of the bpmn definition of a process.
- Added the process list
- Changes to the designer to load the diagram using the nee endpoint.
- A new and simpler process was added to the seeder. Run php artisan db:seed --class=ProcessSeeder 
**Notes**
- The new process form must be added
- The deactivate button is not functional
- The process is edited when pressing the edit rollover icon

Ref: #369


resources/views/processes/index.blade.php
```diff
@@ -0,0 +1,30 @@
+@extends('layouts.layout', ['title' => __('Processes Management')])
+
+@section('sidebar')
+    @include('layouts.sidebar', ['sidebar'=> Menu::get('sidebar_admin')])
```
velkymx: This is the wrong sidebar, but we don't have this one specifically.

resources/views/designer/designer.blade.php
```diff
@@ -1,6 +1,8 @@
 @extends('layouts.layout')
 @section('content')
-    <div id="designer-container"></div>
```
velkymx: Undefined variable: process (View: /home/circleci/builds/feature-369/resources/views/designer/designer.blade.php)

resources/views/processes/index.blade.php
danloa: we changed to  sidebar_designer

resources/views/designer/designer.blade.php
danloa: Now the process is an optional parameter in the controller

velkymx:
Still can't add a start event when I edit a process. Also getting some Vue errors.

image

Issue #406, opened 2018-07-20
Patch API for BPMN


velkymx:
When saving in the designer, send the new BPMN to the processes/{process_id}/bpmn to override it.

PR #416, opened 2018-07-23
Patch API for BPMN, Solves #406

danloa:
- Endpoint  api/processes/{process}/bpmn with http method PATCH added
- Tests for the controller that implements the endpoint added
- The validation of fields of the open_trigger_id,  deleted_trigger_id were changed to allow nulls (it won't always be a trigger for those  operations).


ProcessMaker/Http/Controllers/Api/Designer/ProcessBpmnController.php
```diff
@@ -26,8 +26,23 @@ class ProcessBpmnController extends Controller
...
+     * @return \Illuminate\Contracts\View\Factory|\Illuminate\View\View
+     */
+    public function update(Request $request, Process $process)
+    {
+        $process->makeVisible(['bpmn']);
+        $process->bpmn = $request->get('bpmn');
+        ProcessManager::update($process, [
+            'bpmn' => $request->get('bpmn')
+        ]);
+        return response([], 200);
```
velkymx: We should return a success message or a 204.

*204 No Content represents the request is successfully processed, but has not returned any content.

PR #419, opened 2018-07-25
Create New Process

marcoAntonioNina:
Feature https://github.com/ProcessMaker/bpm/issues/405

- Add Modal to create the process
- Add a template with a process that only has the start element.
- was changed ids of template por new generated uids.
- Add new rest /processes/create

### DIFF
--- a/ProcessMaker/Model/Process.php
+++ b/ProcessMaker/Model/Process.php
@@ -167,7 +167,7 @@ class Process extends Model
      */
     protected $rules = [
         'uid' => 'max:36',
-        'name' => 'required',
+        'name' => 'required|unique:processes,name',
         'process_parent_id' => 'exists:processes',
         'status' => 'in:' . self::STATUS_ACTIVE . ',' . self::STATUS_INACTIVE,
         'create_trigger_id' => 'nullable|exists:triggers,id',
--- a/ProcessMaker/Model/Process.php
+++ b/ProcessMaker/Model/Process.php
@@ -167,7 +167,7 @@ class Process extends Model
      */
     protected $rules = [
         'uid' => 'max:36',
-        'name' => 'required|unique:processes,name',
+        'name' => 'required',
         'process_parent_id' => 'exists:processes',
         'status' => 'in:' . self::STATUS_ACTIVE . ',' . self::STATUS_INACTIVE,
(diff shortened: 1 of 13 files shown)


ProcessMaker/Model/Process.php
```diff
@@ -166,7 +167,7 @@ class Process extends Model
      */
     protected $rules = [
         'uid' => 'max:36',
-        'name' => 'required',
+        'name' => 'required|unique:processes,name',
```
velkymx: Does the name need to be unique? That is not a requirement.

resources/assets/js/processes/index.js
```diff
@@ -1,10 +1,19 @@
...
         filter: ''
     },
-    components: {ProcessesListing}
+    components: {
+        ProcessesListing,
+        ModalCreateProcess
+    },
+    methods: {
+        openModalProcess() {
+            this.$refs.modalProcess.onShow();
```
velkymx: There is a better way to open modals using the built in components. Check out the docs.

https://bootstrap-vue.js.org/docs/components/modal/

resources/views/processes/index.blade.php
```diff
@@ -16,9 +16,10 @@
                                placeholder="{{__('Search')}}...">
                     </div>
                     <div class="col-md-4 d-flex justify-content-end align-items-center col-sm-12 actions">
-                        <a href="#" class="btn btn-action"><i class="fas fa-plus"></i> {{__('Process')}}</a>
+                        <a href="#" @click="openModalProcess" class="btn btn-action"><i class="fas fa-plus"></i> {{__('Process')}}</a>
```
velkymx: Use the built in open method.

resources/assets/js/processes/index.js
marcoAntonioNina: yes, but i need to call function because a rest call to rest api

ProcessMaker/Model/Process.php
marcoAntonioNina: changed.

resources/views/processes/index.blade.php
marcoAntonioNina: yes, but i need to call function because a rest call to rest api

Issue #404, opened 2018-07-20
Remove Unused Designer Options


velkymx:
In process objects menu, update the items to only include these ones. Also remove and corresponding code for removed items. 
![screenshot 2018-07-20 13 07 38](https://user-images.githubusercontent.com/535300/43022974-0d67ec0a-8c1e-11e8-877e-73b17f6b6789.png)

marcoAntonioNina:
solved with https://github.com/ProcessMaker/bpm/pull/420

marcoAntonioNina:
solved with pull https://github.com/ProcessMaker/bpm/pull/420

PR #420, opened 2018-07-25
hide Designer options menu

marcoAntonioNina:
Feature https://github.com/ProcessMaker/bpm/issues/404

- Add option visible in component designer options menu


velkymx:
Closes #404

Issue #409, opened 2018-07-20
Process - Start and Stopping


velkymx:
When I click on the ... icon the icon that shows the "eye", when you click on it, should update the status of the process in the database. If it's set to "stopped" update the column with "inactive" and show are red dot. Also remove the process from the "+ Request" modal.

Issue #399, opened 2018-07-20
API Error Handling


velkymx:
In the event of an API error, call the alert function to display the issue to the end user.

velkymx:
This is merged

PR #421, opened 2018-07-25
Process - Start and Stopping #Solves 409

danloa:
- The deactivate/activate icons were changed to show tha active/inactive state
- View changed to send the deactivate/activate request depending on the current status
- The +Request button shows only the active processes.
Ref: #409


ProcessMaker/Http/Controllers/Api/Requests/RequestsController.php
```diff
@@ -40,4 +44,70 @@ public function index(Request $request)
         // Return fractal representation of paged data
         return fractal($requests, new ApplicationTransformer())->respond();
     }
+
+    /**
+     * Returns the list of processes that can be started by the connected user
```
marcoAntonioNina: Where is it verified that users can create Requests?

resources/assets/js/components/requests/modal.vue
```diff
@@ -79,7 +79,7 @@ export default {
       // Now call our api
       // Maximum number of requests returned is 200 but should be enough
       // @todo Determine if we need to paginate or lazy scroll if someone has more than 200 requests
-      window.ProcessMaker.apiClient.get('processes?filter=' + this.filter + '&per_page=200')
+      window.ProcessMaker.apiClient.get('user/processes?filter=' + this.filter + '&per_page=200')
```
marcoAntonioNina: change per_page by default

resources/assets/js/processes/components/ProcessesListing.vue
```diff
@@ -85,10 +85,36 @@
...
+                    ProcessMaker.apiClient
+                        .put('/processes/' + data.uid, {
+                            status: (data.status === 'ACTIVE') ? 'INACTIVE' : 'ACTIVE'
+                        })
+                        .then(response => {
+                            this.loading = false;
+                            document.location.reload();
+                        })
+                        .catch(error => {
+                            // Undefined behavior currently, show modal?
```
marcoAntonioNina: you could use it 

ProcessMaker.alert(message, type)

to inform you that the state could not be changed.

ProcessMaker/Http/Controllers/Api/Requests/RequestsController.php
danloa: Here we have the problem that we aren't managing assignment rules yet. So for the moment we don't have the documentation. @velkymx  I suggest to create a new ticket to implement this.

resources/assets/js/processes/components/ProcessesListing.vue
velkymx: We're already catching the error. We should not be including empty methods.

caleeli: With #399 the api errors will be shown by default with an alert.

danloa: The catch section of the api calls was removed

PR #423, opened 2018-07-25
API Error Handling. Solves #399

caleeli:
API Error Handling.

image

image


resources/assets/sass/app.scss
```diff
@@ -411,5 +411,5 @@ table.vuetable {
   position: fixed;  
   left:30%;
   right:30%;
-
+  z-index: 1051;
```
danloa: wouldn't it better to have a more meaningful number like 9999 and set it as the maximum to be used for the z-index of all our system?

ProcessMaker/Http/Controllers/Api/Workflow/ActivityController.php
```diff
@@ -14,7 +14,7 @@ class ActivityController extends Controller
     public function complete(Request $request, Process $process, Instance $instance, Delegation $token)
     {
         if ($token->thread_status === Delegation::THREAD_STATUS_CLOSED) {
-            return abort(404);
+            return abort(404, 'Task already closed');
```
velkymx: Let's use 410 Gone
"Indicates that the resource requested is no longer available and will not be available again."

resources/assets/js/management/groups/index.js
```diff
@@ -35,8 +35,6 @@ new Vue({
 
       })
       .catch((err) => {
```
velkymx: Do we need the catch at all?

resources/assets/sass/app.scss
caleeli: Change to sass variable: $alertZIndex

ProcessMaker/Http/Controllers/Api/Workflow/ActivityController.php
caleeli: Fixed

caleeli:
The observations related to the error code, catches and z-index where fixed.

resources/assets/js/management/roles/index.js
```diff
@@ -36,12 +36,7 @@ new Vue({
...
-
-      })
-      .catch((err) => {
-        // @todo Replace with new flashy errors?
-        alert('There was a problem creating the role.')
-      })
+      });
     }
   }
 })
```
danloa: add a line at the end of the file

caleeli: fixed

Issue #373, opened 2018-07-09
Notification Style


velkymx:
Link: https://zpl.io/VYAy7zL

Requirements: 

- Blue text = Task
- Dark gray text = Process
- Dark gray text = Created by 

VIEW ALL TASKS goes to main Assigned Tasks page (https://zpl.io/V41Q5KJ)

<img width="501" alt="screenshot 2018-07-09 15 06 05" src="https://user-images.githubusercontent.com/535300/42478477-b5245462-8389-11e8-9e02-334bab607ee5.png">

meadrat:
@velkymx I updated this with link + requirements

meadrat:
@caleeli @velkymx we need to discuss requirements with Enrique for what's included for Beta.

velkymx:
@milaendo please take this over and clean up the styling for the notification menu.

caleeli:
Solved by #424

velkymx:
See side by side comparison.

![screenshot 2018-08-01 10 25 40](https://user-images.githubusercontent.com/535300/43537859-4ebdfcd4-9575-11e8-98ff-6c5fdd9fe3b4.png)

milaendo:
![screen shot 2018-08-01 at 10 46 22 am](https://user-images.githubusercontent.com/29641725/43538913-29024c36-9578-11e8-9a44-045371737985.png)

milaendo:
![screen shot 2018-08-01 at 10 52 02 am](https://user-images.githubusercontent.com/29641725/43539163-f1ebd41e-9578-11e8-9ca8-770f8b27b6cc.png)

milaendo:
![screen shot 2018-08-01 at 11 03 44 am](https://user-images.githubusercontent.com/29641725/43539740-961f02f8-957a-11e8-993e-b3a853f9b4e9.png)

PR #439, opened 2018-08-01
addressed width and position of the drop down menu 

milaendo:
- also changed font styles and sizes 
#373


milaendo:
screen shot 2018-08-01 at 1 30 37 pm

milaendo:
screen shot 2018-08-01 at 4 06 19 pm

Issue #452, opened 2018-08-06
Process Element Menu


velkymx:
When I click on the "gear" icon in the crown, I'd like to open a menu that will allow me to customize the task. 

![default](https://user-images.githubusercontent.com/535300/43733006-182135aa-9968-11e8-9724-ca8f487ed96a.png)

velkymx:
This is going to change. Hang on.

velkymx:
there are not right click menus

PR #482, opened 2018-08-09
Changes to the Process Element Menu

hjonathan:
Added files to context menu options and add actions to show and open the context menu.
When press context menu in a BPMN Element.

Ref: #452


velkymx:
This is on hold (per the ticket).

PR #493, opened 2018-08-13
Designer Names

marcoAntonioNina:
Feature https://github.com/ProcessMaker/bpm/issues/487

- Add  parameter title


velkymx:
@marcoAntonioNina please update the PR with more details.

Write useful descriptions and titles

Writing a useful description in the “details” section of your pull request can almost be as effective as making a smaller pull request! If I do make a large pull request, I’ll make sure to spend a lot of time making a really helpful description. The most helpful descriptions guide the reviewer through the code as much as possible, highlighting related files and grouping them into concepts or problems that are being solved.

Similarly, a useful summary title (instead of just the issue key - put the key in the description) makes it clear to the reviewer what’s being solved at a high level, which takes off one extra step for the reviewer. At the end of the day, both of these things give the reviewer more context around the change that’s happening.

resources/views/designer/designer.blade.php
```diff
@@ -1,7 +1,7 @@
 @extends('layouts.layout')
 @section('content')
     <div id="designer-container">
-      <designer process-uid="{{$process->uid}}"></designer>
+      <designer process-uid="{{$process->uid}}" process-title="{{ $title }}"></designer>
```
velkymx: Can't we just use $process->title?

ProcessMaker/Http/Controllers/Designer/ProcessController.php
```diff
@@ -32,7 +32,8 @@ public function show(Process $process = null)
             request()->session()->flash('_alert', json_encode(['danger', __('The process was not found.')]));
             return view('processes.index');
         }
+        $title = $process->name;
 
-        return view('designer.designer', compact('process'));
+        return view('designer.designer', compact(['process', 'title']));
```
velkymx: Can't we just use $process->title? Do we need a second variable?

marcoAntonioNina: yes, because in layout.blade.php wait variable "$title" for title of tab.
image

resources/views/designer/designer.blade.php
marcoAntonioNina: actually process-> name and $title have the same value.

"$title" is the title in the tab browser, please  let me know if I should have another text for title

Issue #491, opened 2018-08-12
Generate API Token


velkymx:
We are continuing to have issues with the GenerateApiToken middleware. It needs to be overhauled to be cleaner and more fault tolerant.

Specific issue: Client authentication failed
…/ProcessMaker/Http/Middleware/GenerateApiToken.php97

The PHPUnit tests seem to consistently corrupt the data.

velkymx:
The issue with the Client authentication failed persists. @caleeli Run PHPUnit and then try to login.

velkymx:
Login is getting corrupted after running PHPUnit, make sure your PHPUnit config is set up to use a different database then workflow.

PR #496, opened 2018-08-13
When access token expires, refresh the token and retry the request. #491

caleeli:
When access token expires, refresh the token and retry the request:

image

Ref  #491


velkymx:
@caleeli this doesn't satisfy #491

PR #499, opened 2018-08-14
Add confirm modal in buttons delete

marcoAntonioNina:
Feature https://github.com/ProcessMaker/bpm/issues/485

- was add api delete in class Users, Roles, Groups.
- was add test delete in class Users, Roles, Groups.
- was add action button delete in list Processess, Users, Roles, Groups, forms, inputs and output Document, scripts.


ProcessMaker/Http/Controllers/Api/Administration/UsersController.php
```diff
@@ -103,4 +103,18 @@ public function avatar(Request $request, User $user)
...
+     *
+     * @param User $user
+     *
+     * @return ResponseFactory|Response
+     * @throws \Exception
+     */
+    public function delete(User $user)
+    {
+        $user->delete();
+        return response([], 204);
```
velkymx: `SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`workflow`.`processes`, CONSTRAINT `processes_user_id_foreign` FOREIGN KEY (`user_id`) REFERENCES `users` (`id`)) (SQL: delete from `users` where `id` = 3)`

PR #502, opened 2018-08-14
Flow elements labels Feature/411

hjonathan:
This PR implements designer to show the labels of the BPMN elements.

image


velkymx:
Also getting an error `SQLSTATE[42S02]: Base table or view not found: 1146 Table 'workflow.configurations' doesn't exist (SQL: select * from `configurations` where `parameter` = trustedproxy limit 1)`

Issue #545, opened 2018-08-17
Tasks Grid Update


velkymx:
Update the data grid to match. Make Task Title clickable to the task.

Grid cols: Title, Process, Assigned to, Created by, Due

![screenshot 2018-08-17 10 58 50](https://user-images.githubusercontent.com/535300/44281775-baf9ae80-a20d-11e8-85d1-fc013aecef36.jpg)

caleeli:
@velkymx What does the pause and the undo buttons?
![image](https://user-images.githubusercontent.com/8028650/44343699-ac550680-a45c-11e8-9c40-6ff425902b22.png)

PR #551, opened 2018-08-20
Update Tasks Grid presentation.

caleeli:
- Task title is clickable 
- Introduced User assigned
- Add avatars of users assigned and creator
- Add dot for the due date

image

Ref: #545


caleeli:
Fixed the size and spacing of the avatar and the due date dot

PR #554, opened 2018-08-20
Remove Role CODE attribute.

caleeli:
- Use ID instead of CODE to make the reference to the Roles.
- Use Name as label of the Role
- Updated the unit tests.
- Fixed title of edit popup of the Role.

image

image

image


velkymx:
I'm going to close this. Roles are being removed in #564

Issue #467, opened 2018-08-07
Grid Component Styling


velkymx:
The grid component we're using needs to be updated with the following changes.

- Selected headers should be #333333
- Action buttons are the wrong color #788793
- Status bubbles CSS doesn't match. Use Bootstrap classes
- Change all dates columns to be just the word (created instead of created_on change to created, etc)
- Search box doesn't match mockup

<img width="1163" alt="screenshot 2018-08-07 11 58 24" src="https://user-images.githubusercontent.com/535300/43796457-337fbeb2-9a39-11e8-9f28-a6271559693c.png">

velkymx:
For the circles use the bootstrap pills. 

https://bootstrap-vue.js.org/docs/components/badge#pill-badges

<b-badge pill variant="primary">Primary</b-badge>

KillahDillah:
Search Bar has been updated to match mockup
header color change to match mock up
status bubbles change color according to status type, however not sure this is the preferred method

velkymx:
Regular column (non-selected) the text should be #788793

velkymx:
Please go through all screens and make sure the bubble is updated to the smaller size.

PR #560, opened 2018-08-20
Status bubble colors on Users, Roles, Groups

KillahDillah:
tests pass locally

additional pull request for #467  - original PR #501 
Making necessary changes (accept incoming changes and removed console.log)


KillahDillah:
screen shot 2018-08-22 at 2 09 14 pm

#467 
added 'small' class to TasksList to reduce status bubble size

test pass locally

Issue #535, opened 2018-08-17
New Requests Modal


velkymx:
By default sort by Category Name then Request Name ASC

PR #562, opened 2018-08-21
Updated sorting on New Requests modal

danloa:
When no sort option is sent to the endpoint users/requests (to list the processes that can be started by a user), the list sorted by category name and process name in ascending order.
Tasks done:
- Unit test, to check that the default and custom sorting are working
- Changes to the RequestsController.php to add the default sort behavior when no sortin expression is sent to the endpoint.

Ref #535


danloa:
Now we aren't using full eloquent queries (no Raw commands).

PR #566, opened 2018-08-21
Ability to make a user inactive

milaendo:
- tests passing locally 
- drop down added in edit user modal
- tested manually as well


tests/Feature/Api/Administration/UsersTest.php
```diff
@@ -222,6 +222,7 @@ public function testUpdateUser()
             'username' => app()->make('Faker\Generator')->text(10),
             'firstname' => app()->make('Faker\Generator')->text(10),
             'lastname' => app()->make('Faker\Generator')->text(10),
+            'status' => 'INACTIVE'
```
velkymx: This test needs to be expanded. Test that a user is inactive and then verify that you cannot login with it. Just checking for the 200 is not enough.

milaendo:
@velkymx tests have been written to check that a user cannot login with a status of inactive, but if someone could take over it would be helpful.

milaendo:
- test are passing and we are now checking for the users status before login! thank you @nolanpro for the help :)

Issue #587, opened 2018-08-22
Refactor Auth


velkymx:
In preparation for our new permissions set and the removal of roles, we need to replace the custom authentication with Laravel Auth + Passport. In addition, we need to modify the Auth::attempt to include a check for the status of the user.

PR #591, opened 2018-08-23
Add passport and remove any custom auth behavior. Hooray for Laravel …

tdondich:
…and Passport

Ref #587


tdondich:
To test: Be sure to composer install, as it needs a new laravel package and framework version.

Also, re-run the php artisan bpm:install  . This will ensure the passport encryption keys are installed properly.

Please note, API related tests are now changed.  It can simply be derived from Tests\TestCase and utilize the actingAs($user, 'api') helper to authenticate against the api.  No more manual oauth handshaking.

Go go go!

PR #595, opened 2018-08-23
List Tasks not should display tasks of type script

marcoAntonioNina:
feature https://github.com/ProcessMaker/bpm/issues/580

- Save type of task in table delegation. (will need to define types of tasks to save in the table. )
- list tasks only display task of type normal


velkymx:
@marcoAntonioNina script tasks are still showing on the /tasks page. Also, clicking on the task takes you to an error page instead of the correct task page (probably because it's a script task)

Issue #583, opened 2018-08-22
Closed Requests


velkymx:
From the Requests sidebar, I want to be able to click on the "Closed Requests" and see a list of the requests in the closed/completed status.

PR #606, opened 2018-08-24
Closed requests - allow filtering by status and wire up menus

nolanpro:
#583

Went ahead and added the other statuses to the menu


nolanpro:
@velkymx should be fixed

Issue #527, opened 2018-08-16
About Page


velkymx:
On the admin sidebar, hook up a new page the shows some basic info on the app.

![screenshot 2018-08-16 10 46 43](https://user-images.githubusercontent.com/535300/44225221-bff12c00-a141-11e8-8138-795417603792.png)

velkymx:
Bonus points if you query github and get the latest version from the repo.

PR #628, opened 2018-08-29
About page

milaendo:
screen shot 2018-08-29 at 2 15 32 pm

once we have releases with tags in our github we can update the version placeholder.
#527


webpack.mix.js
```diff
@@ -33,6 +33,7 @@ mix.webpackConfig({
     .js('resources/assets/js/management/queues/index.js', 'public/js/management/queues')
 
     .js('resources/assets/js/management/preferences/index.js', 'public/js/management/preferences')
+    .js('resources/assets/js/management/about/index.js', 'public/js/management/about')
```
velkymx: Delete this

Issue #618, opened 2018-08-28
Task Page Title


velkymx:
Currently, when you click on a task, the page title (<title>) show's Welcome. Should show the name of the task instead. 

![screenshot 2018-08-28 09 46 53](https://user-images.githubusercontent.com/535300/44737630-6543c280-aaa7-11e8-9f9e-281b390fc274.png)

PR #635, opened 2018-08-29
#618 updated blade title to the task name

milaendo:
screen shot 2018-08-29 at 4 39 36 pm

-tests passing locally merged with newest develop
#618


resources/views/tasks/open.blade.php
```diff
@@ -1,4 +1,4 @@
-@extends('layouts.layout')
+@extends('layouts.layout', ['title' => __("{$process->name}")])
```
velkymx: There isn't a translation for this. Just use $process->name

resources/views/tasks/open.blade.php
```diff
@@ -1,4 +1,4 @@
-@extends('layouts.layout')
+@extends('layouts.layout', ['title' => $process->name])
```
velkymx: @milaendo this is the process title, not the "Task" title

Issue #657, opened 2018-10-01
Show User


velkymx:
Add a page to see a user's information without a form. Should be pure HTML. The user's profile picture and details will be listed. 

Path: /admin/user/{id}

HTML: https://github.com/ProcessMaker/ui/blob/master/Admin/userDetail.html

Include UI tests for Route and Page Content

PR #668, opened 2018-10-01
Show user detail template

milaendo:
added show page for userDetail and test for route and content on the page #657


velkymx:
ErrorException (E_ERROR)
Call to a member function format() on null (View: /opt/processmaker/resources/views/admin/users/show.blade.php)
loggedin_at

resources/views/admin/users/show.blade.php
```diff
@@ -9,7 +9,103 @@
...
+        </div>
+      </div>
+    </div>
 @endsection
 
 @Section('js')
+@endsection
+
+@Section('css')
+<link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600,700,800" rel="stylesheet">
```
velkymx: Remove this

tests/Feature/Admin/UserTest.php
```diff
@@ -83,10 +83,10 @@ public function testShowRoute()
       $user_uuid = factory(User::class)->create()->uuid_text;
       // get the URL
       $response = $this->webCall('GET', '/admin/users/'. $user_uuid);
-
       $response->assertStatus(200);
       // check the correct view is called
       $response->assertViewIs('admin.users.show');
+      $response->assertSee('Groups');
```
velkymx: Make this test generate a user and then assertSee the user's name on the page.

resources/views/admin/users/show.blade.php
```diff
@@ -9,7 +9,103 @@
...
+            </tr>
+          </thead>
+          <tbody>
+            @forelse ($user->members() as $group_member)
+            <tr>
+              <td>{{$group_member->group->name}}</td>
+              <td>{{$group_member->group->status}}</td>
+              <td class="text-right"><i class="fas fa-trash-alt text-secondary"></i></td>
+            </tr>
+            @empty
```
velkymx: Remove the groups table entirely if there are not groups found. Put it inside of a card with the text "User is not a member of any groups" centered with padding (See Zeplin).

resources/sass/app.scss
```diff
@@ -68,6 +68,57 @@ html {
   margin: 15px 5px
 }
 
+.avatar-small {
```
velkymx: This is duplicated from the nav bar scope. Can we combine into one?

database/factories/UserFactory.php
```diff
@@ -30,6 +30,7 @@
         'birthdate' => $faker->dateTimeThisCentury->format('Y-m-d'),
         'timezone' => $faker->timezone,
         'language' => 'us_en',
+        'loggedin_at' => $faker->dateTimeThisCentury->format('Y-m-d'),
```
velkymx: Add time to this

resources/views/admin/users/show.blade.php
KillahDillah: removed

Issue #662, opened 2018-10-01
List Environmental Variables


velkymx:
Path: /processes/environmental_variables

HTML: https://github.com/ProcessMaker/ui/blob/master/Processes/environmentalVariables.html

Include UI tests for Route and Page Content

PR #674, opened 2018-10-01
Environmental Variables

milaendo:
#662 testing routes and html brought in


velkymx:
@milaendo this branch is in conflict

Issue #707, opened 2018-10-05
Add User Modal


velkymx:
Enable the add user button in the /admin/users/ page. Show the new user modal, save and go to the /admin/users/{id} page. (show alert)

<img width="539" alt="screenshot 2018-10-10 12 35 21" src="https://user-images.githubusercontent.com/535300/46761335-0b7b0e80-cc89-11e8-9b60-46784191db8f.png">

PR #737, opened 2018-10-11
User list Actions

marcoAntonioNina:
feature https://github.com/ProcessMaker/bpm/issues/710

- add component fields user
- Add Create user with component fields user in window modal.
- Add Edit user with component fields user in route page


ProcessMaker/Http/Controllers/Admin/UserController.php
```diff
@@ -2,33 +2,57 @@
...
+    {
+        return view('admin.users.show', compact('user'));
+    }
+
+    /**
+     * Get all groups actives
+     *
+     * @return mixed
+     */
+    private function getAllGroups()
```
velkymx: What is this method? How does it differ from the index method?

ProcessMaker/Http/Controllers/Api/UserController.php
```diff
@@ -109,11 +111,12 @@ public function index(Request $request)
      */
     public function store(Request $request)
     {
-        $request->validate(User::rules());
         $user = new User();
-        $user->fill($request->input());
+        $user->fill($request->json()->all());
+        //validate model
+        $this->validateModel($user, User::rules($user));
```
velkymx: Why are we validating in a different way? This it not the way we want to do it.

ProcessMaker/Models/User.php
```diff
@@ -107,11 +108,21 @@ public static function rules($existing = null)
     {
         $rules = [
             'username' => 'required|unique:users,username',
-            'email' => 'required|email',
+            'email' => 'required|email|unique:users,email',
+            'password' => 'required',
```
velkymx: Password is not required.

resources/js/admin/users/components/CreateUser.vue
```diff
@@ -1,155 +1,167 @@
...
-                    v-model="value"
-                    :options="options"
-                    :multiple="true"
-                    track-by="title"
-                    :custom-label="customLabel"
-                    label="name"
-                    >
-                    
-                    <template slot="tag" slot-scope="props" >
+        <h1>Create New User</h1>
```
velkymx: Create user modal is a different ticket and is being done in a newer way. Remove this

velkymx: See #707

tests/Feature/Api/UsersTest.php
```diff
@@ -196,14 +196,11 @@ public function testGetUser()
...
   {
-      $faker = Faker::create();
-
-      $uuid = factory(User::class)->create(['username' => 'mytestusername'])->uuid_text;
       //The post must have the required parameters
-      $url = self::API_TEST_URL . '/' .$uuid;
+      $url = self::API_TEST_URL . '/' . factory(User::class)->create()->uuid_text;
 
       $response = $this->apiCall('PUT', $url, [
-          'username' => 'updatemytestusername'
```
velkymx: Does this test work? What are we testing by changing this?

resources/js/admin/users/components/CreateUser.vue
marcoAntonioNina: I'm reviewing feature 707. Thanks.

ProcessMaker/Models/User.php
marcoAntonioNina: ok

tests/Feature/Api/UsersTest.php
marcoAntonioNina: Yes, the test is working. it was proved that the field "username" is mandatory, because this field was sent and only this field should be verified.

the previous test verified that the other required fields were missing.

ProcessMaker/Http/Controllers/Api/UserController.php
marcoAntonioNina: This way he assured me to validate only the fields sent.

for example. change status of user  only send field "status=INACTIVE" if I checking with rules with "validate request" I have the error "Field is required" in (username, etc...)

Issue #727, opened 2018-10-09
Add OAuth login to swagger UI


nolanpro:
Should use a 3-legged OAuth flow so we don't have to generate a personal access token every time we want to try it out.

velkymx:
This isn't working correctly. The OAuth 2.0 3 legged auth should work with the api documentation in order to authorize the api docs. I shouldn't have to include a bearer token.

tdondich:
See: http://oauthbible.com/#oauth-10a-three-legged

PR #755, opened 2018-10-12
Add oauth2 authentication to swagger ui

nolanpro:
#727 

This allows the swagger ui (at /api/documentation) to authenticate using oauth2 instead of personal access tokens. Some additional configuration is needed on local development machines for this to work. See the updates to the README.md to configure


nolanpro:
@velkymx we shouldn't need to ignore it. It's kind of like a lock file. Running artisan `l5-swagger:generate` should fix it. I'll update this branch

Issue #745, opened 2018-10-12
Update VueTables CSS


velkymx:
Update the global stylesheet for the VueTables for the changes from Product. Start with the Users grid on /admin/users and let's review.

<img width="1155" alt="screenshot 2018-10-12 08 55 56" src="https://user-images.githubusercontent.com/535300/46880112-a516fd00-cdfc-11e8-9dc0-1fa4aafd3cc4.png">

velkymx:
https://processmaker.atlassian.net/wiki/spaces/PROD/pages/420675643/Tables+-+Format+Quick+Actions

milaendo:
![screen shot 2018-10-12 at 1 57 29 pm](https://user-images.githubusercontent.com/29641725/46893746-de646280-ce26-11e8-984c-f0c56a2ea9f9.png)

milaendo:
![screen shot 2018-10-12 at 2 00 59 pm](https://user-images.githubusercontent.com/29641725/46893857-43b85380-ce27-11e8-9f40-4f8ad4a1c98e.png)

meadrat:
@milaendo here is the Zeplin link for processes (https://zpl.io/2jwm1n6) and users (https://zpl.io/bevqJXY) so you can crosscheck the design. I see a few minor things. Lmk if you want me to point out or if you're catching them. Thanks!

milaendo:
@meadrat Ill take a look thank you! I am mostly addressing the row hover functionality right now :)

milaendo:
@meadrat I noticed that the vertical line is back in on some of the screens, didn't we agree to drop that?

milaendo:
![screen shot 2018-10-12 at 2 45 09 pm](https://user-images.githubusercontent.com/29641725/46895478-6fd6d300-ce2d-11e8-93f5-7e8174987590.png)

milaendo:
![screen shot 2018-10-12 at 3 06 24 pm](https://user-images.githubusercontent.com/29641725/46896119-6733cc00-ce30-11e8-802b-de823a272b51.png)

milaendo:
![screen shot 2018-10-15 at 2 02 37 pm](https://user-images.githubusercontent.com/29641725/46978157-fdf7c700-d082-11e8-82dc-dfc60e04801a.png)

milaendo:
[...]

PR #769, opened 2018-10-17
Updated global css for data tables

milaendo:
when hovering over tables the action menu will show, all tables that are completed have had ellipsis removed. #745


velkymx:
@milaendo please go through all pages and all screens. Just confirm they're all updated.

milaendo:
@velkymx all but the requests pages have been checked!

PR #771, opened 2018-10-17
Fix group ai

marcoAntonioNina:
- revert list group api
- fix test in rules models


resources/views/admin/groups/fields.blade.php
```diff
@@ -9,7 +9,8 @@
     {!! Form::textarea('description', null, ['id' => 'description', 'rows' => 4, 'class'=> 'form-control', 'v-model' => 'formData.description', 'v-bind:class' => '{\'form-control\':true, \'is-invalid\':errors.description}']) !!}
```
velkymx: Why is this an include? Put it back in the blade. NO INCLUDE

Issue #774, opened 2018-10-17
translate alert messages


milaendo:
go through all axois responses and translate using laravel localization.

tdondich:
merged

PR #791, opened 2018-10-22
axios translations

milaendo:
-in modals with axios responses, all alert responses have been translated
#774


milaendo:
@velkymx tests are passing locally and I am not able to replicate the failing test in ci

PR #799, opened 2018-10-23
Bugfix/793

tdondich:


tdondich:
Closing as 793 is no longer valid as we're no longer using uuids

Issue #857, opened 2018-11-01
Permissions UI


velkymx:
Pulled from the permissions table
Return a list of the permissions
UI will have the name and a checkbox (check if the account has permission)
Save all checkbox items to update user's permissions
Check All
Set as admin select

velkymx:
![image](https://user-images.githubusercontent.com/535300/47882655-767eb780-dde7-11e8-8d93-8de7a08314f2.png)

I think it would be best if we grouped the permissions like this screen.

milaendo:
<img width="1185" alt="screen shot 2018-11-01 at 4 11 19 pm" src="https://user-images.githubusercontent.com/29641725/47885127-cada6500-ddf0-11e8-8076-bf107dce66f8.png">

permissions table placeholder has been added

milaendo:
![screen shot 2018-11-05 at 9 15 51 am](https://user-images.githubusercontent.com/29641725/48014395-6adb0b80-e0db-11e8-9809-2e8bae024307.png)
The way the permissions are set up wont allow for that type of check box list, unless we add a category so we can group them together in that way. @velkymx what do you think about this?

velkymx:
@milaendo This needs to be moved to the edit page /admin/users/1/edit remove it from the profile page entirely

velkymx:
@milaendo this UI doesn't match the requirements. Please refresh to look like the grid.

PR #882, opened 2018-11-09
Permissions UI

milaendo:
first run php artisan migrate:fresh --seed

permissions are located at admin/users/{id}/edit in the permissions tab. This tab is only visible to super users.
If the user already has permissions saved to the db the check boxes will be checked upon load.If the user is a super user in the db, all other permissions will be disabled. 

There has been a new end point added to the api for permissions as well.  It has been annotated and tested. 

There are changes to the User Model and seeder to make the names of the permissions human readable. 

#857


milaendo:
screen shot 2018-11-14 at 2 39 44 pm

velkymx:
This is on hold until #909 is sorted out.

PR #895, opened 2018-11-12
Screen types

milaendo:
in the create modal as well as the show page for individual screens there is now a drop down with the "Display" option. Unit tests have been added for the create method as well as edit.


tdondich:
I instead just had the default screen type be form.  Since this isn't optional, a blank option in the select doesn't make sense.

velkymx:
> I instead just had the default screen type be form. Since this isn't optional, a blank option in the select doesn't make sense.

It should have a SELECT option and validation should pick that up as a failed value

tdondich:
Updated code to have a select option and have validation work against type.

PR #896, opened 2018-11-13
Small Avatar Display

marcoAntonioNina:
bugfix https://github.com/ProcessMaker/bpm/issues/886

- Add style for small avatar display.


resources/js/components/AvatarImage.vue
```diff
@@ -65,14 +66,18 @@
             formatSizeButton(size) {
                 this.styleButton = 'width: ' + size + 'px; height: ' + size + 'px; font-size:' + size / 2.5 +
                     'px; margin-right:5px;';
+                if (size <= 30) {
+                    let s = (size / 2) + 5;
+                    this.styleButton += 'padding-left:' + s + '%';
+                }
```
nolanpro: I don't think this is necessary since the text is automatically centered. I think you just need to add `padding:0` to th styleButton above to solve the problem, at least it worked for me when I tried it

nolanpro: While we're at it, we should probably add `cursor:pointer;` also to indicate it's clickable

marcoAntonioNina: The suggested changes were updated.

marcoAntonioNina:
Add styles in grid with avatar

image

image

PR #897, opened 2018-11-13
Feature/856 - Versions

### LINKED ISSUE #856: Versions
Lean parent tables (id, timestamps)
Models: two methods - latests and versions
API: load the latest version merge with parent return flat array
PARENT: screens
VERSIONS: screen_versions (fk to screen_id)
CREATE will add a top parent record (screen)
CREATE or UPDATE will create version record (screen_versions) and then save the data to the child

NO CHANGES to the UI or AXIOS

Acceptance test:

Test to see if there is a version on create
Test to see if there is more children on update

ENGINEERING NOTES

When you add a hasMany versions function to the model, creating a new version in the controller should be as simple as
```
$model->versions->create($request->all());
```
We can do this in both the update and store functions

For returning the record, we can update the resource. If the controller is calling `ApiResource` directly, create a new resource that inherits it. For example, `class ProcessResource extends ApiResource` then merge the version and it's parent in there. Make sure the ID returned in the payload is the parent, not the version.

nolanpro:
Fixes #856 
For Scripts, Processes, and Screens, this will save a version when the api gets an update request. The versions table mirrors the main table making this pretty easy. No need for a latest() function since the main object will always be the latest.


ProcessMaker/Http/Controllers/Api/ScriptController.php
```diff
@@ -225,9 +225,17 @@ public function store(Request $request)
     public function update(Script $script, Request $request)
     {
         $request->validate(Script::rules($script));
+        $original_attributes = $script->getAttributes();
 
         $script->fill($request->input());
         $script->saveOrFail();
+        
+        unset(
```
velkymx: We should keep the original values for the created_at field.

nolanpro: yeah that makes sense

Issue #918, opened 2018-11-19
Notifications API


velkymx:
We need an API to access the notifications table. Include full CRUD functionality

Acceptance Criteria

Unit test to list notifications
Unit test to get notification
Unit test to add notification
Unit test to change the status of a notification
Unit test to delete a notification
Annotate and Include in the Swagger documentation

PR #926, opened 2018-11-21
Notifications API #918

danloa:
Solves #918 
- Notifications table now have its own model
- get, put, post, delete endpoints for notifications were added
- Tests for every endpoint were added
- Notifications had 2 existing endpoints: one that is used to get the list of notifications of the current user and the other to dismiss notifications. As there were name collision problems with the created endpoints (index and update) they were renamed and the notification UI components were fixed accordingly.


milaendo:
@danloa can you please fix the conflicts in this pr

Issue #909, opened 2018-11-16
Permissions Mess


velkymx:
The permissions are a bit all over the place and need to be cleaned up. When working on these types of integrations we need to be more careful in our implementation - thinking about how it is going to actually be used rather than blindly adding values.  If there are things we're unsure about you can always reach out to me to discuss and clarify.

1. _Store and Create are the same thing_
2. _Edit and Update are the same thing_
3. _Process_Categories is missing edit and create_
4. _Files is a helper of different parts of the site and doesn't require restrictions_
5. _Requests permissions are determined by the process_
6. _About page should never be restricted_
7. _Process Events Trigger needs to be removed this is part of the process_
8. _Script Preview is part of the Script Edit (and therefore required to edit)_
9. _Tasks (show and update) should be removed, they're part of the process permissions_
10. _Group Members is missing create and edit_
11. _Preferences is only edit_
12. _Menu access doesn't make sense. It should be based on if the user has access to any of the pages in that area. We can probably kill this with UX/UI_
13. _Where is the list permission (turn off the left sidebar item and hide the grid)?_
14. _Horizon permissions are missing_
15. _Screens permissions are missing_
16. _The All Request list should be restricted_

milaendo:
@velkymx Horizon is only accessible under the admin tab, do you think it should still have its own permissions?

tdondich:
If you can get to something by url then it's not really protected.

nolanpro:
I suggest we make the seed user have all permissions instead of is_administrator. That way when we add new endpoints the developer will have to add the permission

velkymx:
The is_admin is the quick way to get full access to the site. If a new end point is added the developer should have to add the permission explicitly.

PR #932, opened 2018-11-26
Permissions clean up

milaendo:
#909 
About page has been removed
all menus have been removed
Store and Update have been removed (duplicate of create and edit)
Files have been removed
Requests have been removed
process_categories now has edit and create
process event triggers have been removed
Tasks have been removed
Group_members now have edit and create
preferences is now just edit
screens permissions have been added (create, destroy, edit, show)
new permission for show all requests has been added


ProcessMaker/Models/ProcessRequest.php
```diff
@@ -266,4 +267,20 @@ public function summary()
...
+
+    /**
+     * Check if the user has access to this request
+     *
+     * @param User $user
+     * @return void
+     */
+    public function authorize(User $user)
+    {
+        if ($this->user_id == $user->id || $user->is_administrator) {
```
danloa: use === instead of == to avoid implicit casting

PR #933, opened 2018-11-26
Permissions UI

milaendo:
screen shot 2018-11-26 at 10 26 41 am

- under the admin 'users' section there is now a tab for permissions
- If the user is a 'super admin' all other permissions will be disabled
- If the user has permissions already the checkboxes will be pre selected
#857

### DIFF
--- a/resources/views/admin/users/edit.blade.php
+++ b/resources/views/admin/users/edit.blade.php
@@ -109,7 +109,7 @@
                                     <th><input type="checkbox" v-model="isAdmin"></th>
                                 </tr>
                                 <tr>
-                                    <th>Permissions</th>
+                                    <th>Select All Permissions</th>
                                     <th><input type="checkbox" v-model="selectAll" @click="select" :disabled="isAdmin = isAdmin"></th>
                                 </tr>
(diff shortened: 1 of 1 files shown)


ProcessMaker/Http/Controllers/Api/PermissionController.php
```diff
@@ -0,0 +1,96 @@
...
+                    PermissionAssignment::where([
+                        'permission_id' => $permission->id, 
+                        'assignable_type' => User::class, 
+                        'assignable_id' => $user->id
+                    ])->delete();
+                }
+            }
+        }
+
+        return response([], 204);
```
danloa: status 204 means no content returned, better to use:  return response('',204)

milaendo: this is a safe response because you cannot have an empty Json (it will not be valid at that point). We are using this is many other places as well. please talk with @nolanpro if you disagree :)

velkymx:
@nolanpro this PR should not have been merged. It did not fulfill the UI requirements.

Issue #921, opened 2018-11-19
Cancel Request Permission


velkymx:
From the route /processes/{id}/edit add a dropdown to select a group or user who can cancel that request. Once the user/group has permission the request cancel will be accessible via the Request Detail page (see #916). Admins are excluded from this restriction.

Acceptance Criteria

Unit Test: Update process and confirm the api doesn't allow cancelling without permission
Unit Test: Update process and confirm the api does allow cancelling with permission
Unit Test: Confirm the permission assignment with the process in the API call

Issue #920, opened 2018-11-19
Start Request Permission


velkymx:
From the route /processes/{id}/edit add a dropdown to select a group or user who can start that request. Once the user/group has permission the request will be accessible via the New Request
modal. Admins are excluded from this restriction. 

Acceptance Criteria 

Unit Test:  confirm the /api/1.0/processes?include=events,category doesn't return results without permission

Unit Test:  Give the user permission and confirm the /api/1.0/processes?include=events,category does return results with permission

Unit Test:  Admin users only confirm the /api/1.0/processes?include=events,category does return results without permission

PR #934, opened 2018-11-26
Cancel and start Request Permissions #921, #920

danloa:
This PR solves issues #921 and #920

- Permission 'requests.cancel' created
- To create a request the permission to use is 'requests.store'.
- To store the cancel/create requests permissions, a new table "process_permissions" has been created.
- The end point index of the processes was modified so that it returns the list of processes to which the user has access to.
- The user administrator doesn't have any restriction for canceling or allowing to create requests.
- Tests were added


ProcessMaker/Http/Controllers/Api/ProcessController.php
```diff
@@ -54,18 +59,39 @@ class ProcessController extends Controller
...
+            });
+
+            $processes->orWhereExists(function($query) use ($startPermission){
+                $query->select(\DB::raw(1))
+                    ->from((new ProcessPermission)->getTable())
+                    ->whereRaw("process_permissions.process_id = processes.id " .
+                        " AND permission_id=" . $startPermission  .
+                        " AND assignable_type = '" . str_replace('\\', '\\\\', Group::class) . "'" .
+                        " AND assignable_id IN (select group_id from group_members where member_id = " .
+                        Auth::user() ->id . " )");
```
nolanpro: I think there's probably a cleaner way to do this. We should try to avoid any raw sql when ever possible. I think the logic should get moved out of the controller and into the [chainable scope](https://laravel.com/docs/5.7/eloquent#local-scopes) on the process model, something like `forUser($user)`

PR #935, opened 2018-11-26
Update .env and Laravel Echo Server configs

daniel-ryan-PM:
Add redis_port and placeholder in .env file. Gets updated with correct port via deploy script.
Also added bpm_scripts_home tmp directory.
Change echo server to http and add ECHOPORT placeholder.


tdondich:
Change identifiers to replace with %%VARNAME%%

Copy laravel-echo-server.json to the .circleci directory.  Modify that all you want.  On deploy, copy that file and replace the one at the root dir for deployment.

PR #958, opened 2018-11-29
Upload files to a request and be able to download them

milaendo:
screen shot 2018-11-29 at 3 29 46 pm

- the screen builder now has a new type called "file upload"
- when a file is saved it is uploaded to the media library and a reference id is saved to the json of the request
- when a request has had all tasks completed there will be a new tab in the summary screen for attached files.
- each file name is a link to download the file


resources/views/requests/show.blade.php
```diff
@@ -88,6 +92,30 @@
                             <request-detail ref="completed" :process-request-id="requestId" status="CLOSED">
                             </request-detail>
                         </div>
+                        <div class="tab-pane fade" id="files" role="tabpanel" aria-labelledby="files-tab">
+                        <div class="mt-3">
+                            <div>
+                                <table class="vuetable table table-hover">
+                                    <thead>
+                                        <tr>
+                                            <th>File Name</th>
```
marcoAntonioNina: use __('text') for labels

PR #992, opened 2018-12-04
As a user without permission to cancel should not see the cancel button

### LINKED ISSUE #976: Cancel Request
The cancel request button should only show up if the user has permission to cancel that request in the process permissions.

milaendo:
fixes #976 

To test: 
log in as a user without permission to cancel a request, navigate to the request summary page and see that there is no long a section for "cancel"

There has also been unused code removed in the ProcessRequestController
screen shot 2018-12-04 at 3 32 46 pm


ProcessMaker/Http/Controllers/RequestController.php
```diff
@@ -61,8 +61,9 @@ public function show(ProcessRequest $request, Media $mediaItems)
         $request->user;
         $request->summary = $request->summary();
         $request->process->summaryScreen;
+        $canCancel = Auth::user()->hasProcessPermission($request->process, 'requests.cancel');
```
velkymx: Does this catch for admins? If is_admin then the check should always pass.

resources/views/requests/show.blade.php
```diff
@@ -131,6 +131,8 @@
                                 <avatar-image size="32" class="d-inline-flex pull-left align-items-center"
                                               :input-data="requestBy" display-name="true"></avatar-image>
                             </li>
+                            
+                            @if($canCancel == true)
```
velkymx: Is there vue code to check for as well?

PR #1014, opened 2018-12-11
Vuetable CSS Refresh

### LINKED ISSUE #1010: Vuetable CSS
- [x] Remove the custom styling and use the built in Bootstrap styling for the vue table component 
- [x] Remove the hover effect and put the action icons in the last column
- [x] Action buttons should always include an open icon (instead of the column you click on)
- [x] Use consistent icons

milaendo:
fixes #1010 

There has been a few factories updated in this ticket (populating with the wrong kind of data)
All data tables have been updated with new classes and icons
Pagination is now only visible when there is more than one page of info
Font sizes have been updated so that they are not so small now
Old classes have been removed from app.css


velkymx:
Go through each screen and make sure buttons, status and dates don't wrap.

PR #1019, opened 2018-12-12
File Downloader/ File Uploader

### LINKED ISSUE #989: File Upload Enhancements
[]- In the event a file is uploaded, and a file already exists due to an uncompleted task, that old file should be deleted and replaced with the new one.
[]- Add download link to the screen designer that will pick up the id and show that link to download the file
[]- Verify multiple file upload widgets work for the same task (and you can access all of those files)

milaendo:
fixes #989 

this ticket has resolved : 
- the styling issue that was happening when more than one uploader was brought onto the canvas.
- made a new control for the file downloader in both the "form" and "display" screens
- deletes orphaned files
- also fixed the Process request factory so it returns an object not a string


velkymx:
@milaendo use button class secondary (grey) icon file-download
screenshot 2018-12-13 10 47 03

nolanpro:
@velkymx should be good to go

ryancooley:
@milaendo and I added the ability to attach a name to a file downloader component in order to display only uploaded items which correspond with that name.

screen shot 2018-12-20 at 11 10 55 am
screen shot 2018-12-20 at 11 11 05 am
screen shot 2018-12-20 at 11 11 09 am

Issue #1001, opened 2018-12-06
Remove Task/Request Route Permissions


velkymx:
Tasks and Requests as a route permission should be removed. Tasks are assigned to use via the request. No need for a route permission. I should ALWAYS be able to see requires and tasks as long as the process grants me permission to do so.

PR #1027, opened 2018-12-14
Remove of route permissions for task and request controllers

danloa:
Solves #1001 

Route permissions are ignored in Tasks and Requests controllers


nolanpro:
@danloa It's OK to skip these permission checks if there are additional checks in the methods. For example, `ProcessRequestController->index` checks if the user is admin or the task is `startedMe`. We would need similar checks on the other routes where we are skipping permissions. For Tasks, I guess it would be the assigned user (or admin). For ProcessRequests, it's probably if they are a participant (or admin)

Issue #993, opened 2018-12-05
Archive Process


velkymx:
We need to add a way to delete a process. I'd like to be able to archive them, turn them off and move them to a different screen. You cannot start archived requests and it should not show up in the New Requests menu

- [x] Add soft delete
- [x] Add archive list (with restore process)
- [x] Add to the sidebar an archive button

velkymx:
Restore fails with 404 on /api/1.0/processes/1/restore

PR #1050, opened 2018-12-17
Archive Process

danloa:
Solves #993

- Trait to implement soft delete was added
-  Processes listing can be restricted by the process status value
- Tests were added to validate the processes' soft delete functionality.


velkymx:
@danloa you have conflicts and restore does not work. 404 https://github.com/ProcessMaker/bpm/issues/993#issuecomment-448326703

Issue #1046, opened 2018-12-17
Edit User Copy


velkymx:
Above the password fields add the text, "Leave the password blank to keep the current password"

PR #1069, opened 2018-12-19
Edit user text over password fields

danloa:
Solves #1046 

The text was put above the password fields

usrpswd


velkymx:
This is already fixed

Issue #1060, opened 2018-12-18
User Assignment to Other Users


velkymx:
Fix the user assignments. Currently only processes with tasks assigned to 'admin' work properly. It is important to have tasks assignable to other users and have those work when logged as those users.

PR #1081, opened 2018-12-20
Feature/1055 - Keep task assignments after node name change and refactor

### LINKED ISSUE #1055: Renaming a Task Node
when renaming a task node identifier the assigned user is reset (it's not updating the node id in the process_task_assignments table)

nolanpro:
Fixes #1055 

NOTE: this PR includes includes merged #1060, so that one should be merged to develop first.

This refactors the task assignment component to remove unused parts (multi-assignment) and cleans up how we handle task assignment updates. When the node name changes, it triggers an update.

New issue: changing a node name triggers a task assignment update in real-time, however, if the user never saves the BPMN by clicking the save icon, the new node name won't be referenced in the BPMN, and the task assignments will become invalid.

Possible solutions:
- only save task assignment updates when the user clicks the save icon, not real time like we do now
- reference task assignment ID(s) in the BPMN


danloa:
I've found a minor issue:
- Open any process with at least 2 tasks that we'll call as task A, task B.
- In "Task A" set the assignment to requester
- Assign "Task B" to user admin.
- Save the process.
- Reload the process.
- Select Task A
- Now select Task B
==> The configured user (admin) is not loaded for Task B. If at this moment the user saves the process, the Task B won't have an user assigned.

nolanpro:
Heres the problem. When switching between nodes in the modeler by clicking on them, the mounted() event is not fired on the inspector extension component, in this case `TaskAssignment.vue`. It does, however, fire when the user clicks on a blank area first, then on another node, but we can't expect them to do that. So when switching nodes, fetching the assignments (a sperate axios call), is never run, and the assigned user not set on the dropdown.

From inside the context of TaskAssigment.vue, we have no way of knowing if a user clicked on another node. We can "watch" the node name for changes, however, the user could just be editing the node name, which would also trigger a watch change. In fact, we count on that watcher to update the assignments node name when the user did intentionally change the node name. Because of that, when ever a user clicks on another node, technically the node name changed, so the watch is triggered and a rogue assignment update api call is sent.

The question is, how can the inspector extension (TaskAssignment.vue) know when a user has switched nodes? The Inspector is technically always the same component (an instance of vue-form-renderer), only its config gets updated when the user clicks on another node. We can not "watch" for changes on the parent (like the parent's config).

One option would be to modify the modeler and vue-form-editor libraries to send the config down as a "prop", which can then be watched. Another option is to trigger an event, maybe on the global event bus, alerting that the user has clicked on another node. There may be another solution that @tdondich can come up with.

One option is to send the config as a prop by modifying the vue-form-renderer library

caleeli:
@nolanpro that's right, when you switch between elements of the same type, the inspector will not be re-rendered (since its config is the same) but yes the data, it will change, you will see that the properties correspond to those of the task you have selected.

PR #1083, opened 2018-12-20
Implement Tab to see Request runtime errors

caleeli:
In this PR the following changes were implemented:

- A tab to show the request errors
- Add a json column to store the runtime request errors
- Update the vue-form-builder version adding the default values for the controls

image

image


tdondich:
What is the screenshot of the modeler showing?

tdondich:
And is viewing the errors permission based?

caleeli:
The modeler is showing the process that @josemaldonad0 reported in the ticket when the gateway failed. It was caused because an undefined variable in the condition expression (Well that was caused because and issue fixed in the form builder).

image

The Error tab will show request runtime errors in expressions (Ex. use of undefined variables, wrong data types, ...) and scripts (Exceptions and errors during its execution).
The example is showing an error introduced on purpose in a task script.
```
<?php
throw new Exception('Exception in the script');
```

PR #1120, opened 2019-01-02
Ignore route authorization for notification updates

### LINKED ISSUE #1109: Notifications wont clear for users other than admin
As a user not admin, when a task is completed the notification pane does not clear/dismiss. It still displays the bell count and link to said task. When clicked, it goes to 'Task Completed'. but the notifications do not clear.

nolanpro:
Fixes #1109

- IDs are guids so we shouldn't need to authorize the user for updates since they would be hard to guess
- Also fixes the read/unread in the list of notifications (clicking on the left mail icon) by returning a promise

### DIFF
--- a/resources/js/notifications/components/NotificationsList.vue
+++ b/resources/js/notifications/components/NotificationsList.vue
@@ -82,13 +82,20 @@
         },
         methods: {
             read(id) {
-                ProcessMaker.removeNotifications([id]);
-                this.fetch();
+                console.log("loading..");
+                ProcessMaker.removeNotifications([id]).then(() => {
+                    console.log("done");
+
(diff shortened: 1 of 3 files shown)


resources/js/notifications/components/NotificationsList.vue
```diff
@@ -82,13 +82,20 @@
         },
         methods: {
             read(id) {
-                ProcessMaker.removeNotifications([id]);
-                this.fetch();
+                console.log("loading..");
```
velkymx: Do we need this console logging?

nolanpro: Oops, yup will remove

Issue #1110, opened 2018-12-28
Notifications showing current time, not actual time of task/notification


josemaldonad0:
When clicking on the notification icon, all notifications show **now**, or current,  as the same datetime, rather than the datetime when the notification was triggered. As in, not showing when the task was assigned but rather showing when the notification pane was drawn.

PR #1122, opened 2019-01-02
Notifications showing current time, not actual time of task/notification

danloa:
Solves #1110 

- Notification Controller was modified so that now it sends the dateTime value of the notification's dateTime attribute.
- The notification list was modified to use the dateTime value.


velkymx:
@danloa I dont think this change is necessary. What does the dateTime attribute offer that created_at doesn't?

PR #1150, opened 2019-01-04
Package Manager for Enterprise Capability

MiltonPM:
# The Idea
The package listing api controller fire off a PackageEvent class which all packages listen for. The event has a reference to a package manager (This a laravel's collection ), in this collection is added every installed package's information. Into each package there's a PackageListener class and each listener, which is in each package's PackageServiceProvider's boot method. Once all listeners are completed, then the controller can render the api response data accordingly.

# The musts
- [x] Add page to admin section that lists currently installed packages. This data should be included:
- [x] Package composer name
- [x] Package friendly name
- [x] Package description
- [x] Package expiration (if the package is encrypted, and there is an expiration date of when it will execute)
- [x] The grid that shows this data should be the same data grid control we use everywhere else
- [x] The search should be live search just like we have everywhere else
- [x] The data to fetch should follow our REST API standards.
- [x] The expiration date that is shown should follow the user's timezone and date/time format display.

# What has been done

The related route, Controller, Event and Front end respecting the current development line.


ProcessMaker/Events/PackageEvent.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Get the channels the event should broadcast on.
+     *
+     * @return \Illuminate\Broadcasting\Channel|array
+     */
+    public function broadcastOn()
+    {
+        return new PrivateChannel('channel-name');
```
caleeli: The broadcast channel name looks undefined (or using the default value 'channel-name') This should returns the channel that the event should broadcast on. Examples:

- `ProcessMaker.Models.User.123`  ==>   `channel for user 123 only`
- `notifications` ==> `A channel named notifications`

Could be useful to define a proper channel name for this event, and define how are going to listen it.

Issue #1153, opened 2019-01-07
Overdue Tasks


velkymx:
As a user, when I visit the Tasks area, I want to see a message that let's me know I have overdue tasks that need my attention. 

- [x] If there are overdue tasks, show alert with "You have X overdue tasks pending"
- [x] Include a unit test to confirm that the modal shows correctly

PR #1176, opened 2019-01-11
Overdue Tasks #1153

danloa:
- Tasks list now shows an alert message with the number overdue cases of the logged user
- The number of overdue cases is returned in the meta property of the endpoint.
- Added dusk test to validate that the alert message.
overdue3


danloa:
@velkymx , are you using and admin account? To count the overdue tasks, I'm filtering the tasks that the logged user owns. The administrator can see the tasks of all users, so it will be a difference between this list and the number of tasks assigned to the administrator. Do you think that we need to count the number of overdue tasks differently?

PR #1187, opened 2019-01-15
Add auth client management to admin

### LINKED ISSUE #816: Passport Client Access
During the install process, create an oauth client that can be used for API Documentation.  Then ensure the configuration for the api documentation uses the oauth client created.  This should help facilitate the 3 legged oauth and not prompt the user to have to put in the oauth client details when they authenticate the api docs.

image

image

**Reference**

https://mattstauffer.com/blog/introducing-laravel-passport/
https://laravel.com/docs/5.7/passport#managing-clients

- [x] List client tokens
- [ ] Unit test token works/fails and can be created / regenerated

nolanpro:
Fixes #816 

Adds a section and menu item to admin for creating auth clients. Each of these auth clients will allow 3rd party apps to log in using any of our users credentials.

image


velkymx:
The permissions had a bug with a comma missing

Issue #1183, opened 2019-01-14
Logging Changes


velkymx:
When a user makes a change to a task data or request data I want to log that the data is changed. We are going to use this as the foundation for the comment system. 

<img width="746" alt="screenshot 2019-01-14 10 14 44" src="https://user-images.githubusercontent.com/535300/51131664-3e332680-17e5-11e9-9076-26b0a401791b.png">

- [ ] Add table: comments (id, user_id, poly, subject, body, hidden, created_at, updated_at)
- [ ] Add model: comment
- [ ] Comments should be polymorphic (to be used with tasks, requests, messages, etc)
- [ ] Automatically add comment when data is changed
- [ ] Automatically add comment when task is submitted
- [ ] Unit test to confirm log comments writted

velkymx:
Format ideas:

(avatar) TIMESTAMP USERNAME LOGMESSAGE
(AB) 01/16/2019 7:12AM Admin canceled the request
(AB) 01/16/2019 7:12AM Admin changed the data for Approval task

![screenshot 2019-01-16 06 56 53](https://user-images.githubusercontent.com/535300/51258201-44004780-195e-11e9-9b9c-81ccad158af6.png)

velkymx:
Use the avatar component and link to the user's profile.

danloa:
@velkymx  In the comment system, the table has a body and a subject for every comment. In the line:
(avatar) TIMESTAMP USERNAME LOGMESSAGE
LOGMESSAGE = subject
and with a mouseover the body of the message should be displayed?, Do you agree?

velkymx:
No. No mouse over. ALSO show the logs for tasks in the Request detail under the tabs. So if a task is change (data, status, etc) the log on the Request detail will update

Layout: (AA) TIMESTAMP USERNAME BODY

velkymx:
![screenshot 2019-01-18 14 09 05](https://user-images.githubusercontent.com/535300/51415871-339dc780-1b2c-11e9-8851-0c8d696580bb.png)

Issue #1180, opened 2019-01-11
Process Permissions


velkymx:
Process Permissions - are permission assigned to a user via a task or process configuration.

- [x] Canceling a specific process
- [x] Starting a specific process
- [x] Doing an assigned task
- [x] Reassigning an assigned task
- [x] Editing Data

Acceptance
- [x] Decouple Process permissions from UI permissions 
- [x] Set up process permissions
- [x] Add test to verify permissions are respected
- [x] Add permission checks to navigation

NOTE: Requests area - by default I can see only requests I have access to via my participation. The only route that is different is the All Requests area (/requests/all) where all requests are available.

Issue #1175, opened 2019-01-11
Edit Data


velkymx:
As a user I want be able to edit the data set as a process flows. 

For tasks that have not been completed, add a tab that has key / value pairs of the Vue data that can be edited. 

From the Request Summary, if the request is completed,  add a tab that has key / value pairs of the Vue data that can be edited. 

- [x] Add to config UI group / users who can edit data (allow multiple)
- [x] Add tab to task screen that shows the data (maybe JSON with the Script editor?)
- [x] User can edit data of a task ONLY if the task is status "In Progress"
- [x] User can edit data of Request ONLY for requests in status "Completed"
- [x] Unit test that this permission is being respected
- [x] Verify that when an edit happens a comment is generated (event) See #1183

caleeli:
@velkymx how the data should be edited, are we going to use a JSON editor?
Remember that the data could be saved by any previous screen and could contain complex data like arrays of objects, calculated fields, texts, numbers and booleans. Also the scripts and connectors could save data with any json structure.

caleeli:
Request:
![image](https://user-images.githubusercontent.com/8028650/51190953-c3880b00-18b9-11e9-8c50-42b2f2fbd648.png)

Task:
![image](https://user-images.githubusercontent.com/8028650/51192073-25497480-18bc-11e9-9db8-4dbca2241580.png)

ryancooley:
@caleeli As part of ticket #1180, we are now using [Laravel Policies](https://laravel.com/docs/5.7/authorization#creating-policies) to determine user permissions for various models.

In our branch (feature/1180-b), we have created an update method to the Process Request policy. You can access this in several ways.

- From within a controller, you can add `$this->authorize('update', $processRequest);`
- Via the user model, you can check the truth of the statement `$user->can('update', $processRequest)` 
- Via a Blade template, you can use `@can('update', $processRequest)` and its closing tag `@endcan`

If you have any questions, please feel free to ping @milaendo or myself, or we can discuss during the standup.

velkymx:
@caleeli Let's just dump the JSON to the Javascript editor (the one from the scripts area).

nolanpro:
Merged

PMQABen:
✅ Add to config UI group / users who can edit data (allow multiple)
✅ Add tab to task screen that shows the data (maybe JSON with the Script editor?)
✅ User can edit data of a task ONLY if the task is status "In Progress"
[...]

PR #1200, opened 2019-01-18
Logging Via Comments, #1183

danloa:
- Comments are shown in the request summary
- A comment is added every time that a task is completed
- Migrations were modified to include the new polymorphic tables.
- Tests were added

NOTES:
* In this ticket we aren't adding a comment whe the data is modified because we don't have this functionality. This feature is implemented in #1175 and in this ticket those comments will be added.
* We are using the current infrastructure of permissions, so we'll need some tweaks after the #1180 (Process Permissions) is finished. 

comments3


velkymx:
screenshot 2019-01-18 14 09 05

Issue #1191, opened 2019-01-16
500 Server Error when adding new users with no status set


PMQABen:
1. Go to `/admin/users`
2. Click `+USER`
3. Fill out the fields to add a new user, but leave "Status" blank
4. Save

Result: `500 Server Error`
```
[Vue warn]: Error in render: "TypeError: Cannot read property 'username' of undefined" -- vendor.js?id=820bccc06abf04eb00e9:53823
(found in <Root>)


TypeError: Cannot read property 'username' of undefined -- vendor.js?id=820bccc06abf04eb00e9:54971
    at Proxy.eval (eval at createFunction (vendor.js?id=820bccc06abf04eb00e9:63910), <anonymous>:3:538)
    at Vue._render (vendor.js?id=820bccc06abf04eb00e9:57774)
    at Vue.updateComponent (vendor.js?id=820bccc06abf04eb00e9:56018)
    at Watcher.get (vendor.js?id=820bccc06abf04eb00e9:56372)
    at Watcher.run (vendor.js?id=820bccc06abf04eb00e9:56449)
    at flushSchedulerQueue (vendor.js?id=820bccc06abf04eb00e9:56211)
    at Array.<anonymous> (vendor.js?id=820bccc06abf04eb00e9:55067)
    at flushCallbacks (vendor.js?id=820bccc06abf04eb00e9:54988)


[Vue warn]: Error in render: "TypeError: Cannot read property 'password' of undefined" -- vendor.js?id=820bccc06abf04eb00e9:53823
found in
---> <VuePassword>
       <Root>


TypeError: Cannot read property 'password' of undefined -- vendor.js?id=820bccc06abf04eb00e9:54971
    at fn (eval at createFunction (vendor.js?id=820bccc06abf04eb00e9:63910), <anonymous>:3:3925)
    at Proxy.renderSlot (vendor.js?id=820bccc06abf04eb00e9:56980)
    at Proxy.render (index.js?id=2231f1b0df25c70362a4:4225)
    at VueComponent.Vue._render (vendor.js?id=820bccc06abf04eb00e9:57774)
    at VueComponent.updateComponent (vendor.js?id=820bccc06abf04eb00e9:56018)
    at Watcher.get (vendor.js?id=820bccc06abf04eb00e9:56372)
    at Watcher.run (vendor.js?id=820bccc06abf04eb00e9:56449)
    at flushSchedulerQueue (vendor.js?id=820bccc06abf04eb00e9:56211)
[...]

PR #1201, opened 2019-01-18
500 Server Error when adding new users with no status set #1191

marcoAntonioNina:
Bugfix https://github.com/ProcessMaker/bpm/issues/1191

- Add rule status in Model User
- fix test user with status.


PMQABen:
Fix works as expected. 👍

Issue #1198, opened 2019-01-17
Adding user to group does not update page unless reloaded


PMQABen:
1. Create new group
2. Edit the new group
3. Go to "Group Members"
4. Click +User, select one or more users, press save

**Result:**
The "Group Members" tab does not update to reflect the new user(s) until the page is reloaded, and you navigate back to the "Group Members" tab.

**Expected:**
The "Group Members" tab updates once the save button is pressed, to show the newly added user(s).

nolanpro:
This has been merged. Closing

PR #1202, opened 2019-01-18
Adding user to group does not update page unless reloaded #1198

marcoAntonioNina:
Bugfix https://github.com/ProcessMaker/bpm/issues/1198

- Fix reload list Group members


PMQABen:
Fix works as expected. 👍

PR #1248, opened 2019-01-24
Implements Edit of Request Data #1175

caleeli:
This PR includes:

- Add a config UI group / users who can edit data (allow multiple users and groups)
- Add tab to task screen that shows the data
- User can edit data of a task ONLY if the task is status "In Progress"
- User can edit data of Request ONLY for requests in status "Completed"
- Unit tests to validate the permission of edit data
- When an edit happens a comment is generated (related to ticket 1183)

Related to: #1175


caleeli:
@nolanpro the fixes were committed.

PR #1308, opened 2019-01-31
Set echo server settings in dotenv and update installer to add them

nolanpro:
Fixes #1127 
- Remove the hard-coded configs from laravel-echo-server.json and use .env variables instead
- Update the installer to set those variable
- Restart the echo server (and other services) when the install script finishes so it picks up the new changes
- Update readme with socket.io info. You do not access the socket.io js file manually and allow "unsafe" connection it will not be allowed by the browser because it runs on a different port


nolanpro:
Verified with @Kookster310 the qa server is deploying correctly with these settings.

Issue #1311, opened 2019-02-01
Default Task Assignment to Requester


velkymx:
When I load a task without an assignment, the form should default to the requester user.

- [x] Update inspector to have a default value (and assignment rule)

PR #1325, opened 2019-02-04
Set default "requestor" as default assignment

caleeli:
This PR includes:

- Set "requestor" as the default assignment rule for tasks

image

Related: #1311


PMQABen:
Works as expected.

Issue #1309, opened 2019-02-01
Assign to Previous Task Owner


velkymx:
I want to setup my task to be assigned to the previous tasks assignee. If the previous element is a script or gateway, then go back further to get the task which is assigned to a user. NOTE: We are not evaluating the previous tasks assignment rules. These have already been processed and a singler user has been assigned. 

- [x] Add "to previous task assignee" assignment rule in the drop down
- [x] Before a task is assigned, search the tokens table for a previously assigned task and use that users id for the new assignment.
- [x] Include a unit test to confirm the use case

PR #1326, opened 2019-02-04
Add Previous Task Assignee Rule

caleeli:
This PR includes:

- Add an option in the Modeler to use "Previous Task Assignee Rule" in the tasks
- Add the Previous Task Assignee Rule
- Add unit tests to validate this rule.

image

Related: #1309


rcgongora:
What if we have this process?
image
Task1 is assigned to Admin admin, then we have a Script task(which is not assigned to anyone) then a Gateway and finally two tasks(task2 and task3 which have 'Previous task assignee') both of them should be assigned to Admin admin, but they aren't.
How should behave this assigment rule in this case?

caleeli:
@rcgongora, thanks for the feedback, yes the script task should be skipped to get the Task1 assigned user (fixed), but since it is a _Exclusive Gateway_ only one of the tasks (task2 or task3) will continue.

Issue #1310, opened 2019-02-01
Assign to User by Eval


velkymx:
In the inspector of a TASK ONLY, add the ability to assign users in special cases. 

Under the current assignment rule (this is the default rule)

- [x] Add a new section that is called "Special Assignments"
- [x] Add a button that hide / show special assignment form
- [x] Add form that shows the FEEL evaluation and then THE SAME assignment drop down (to requester, to user, to group, etc) 
- [x] Update assignment to check if there is a special assignment always that will get applied only if the FEEL eval is true. If the eval is false, use the default rule.
- [x] Allow process designer to delete special assignments (no edit)

![screenshot 2019-02-01 12 45 18](https://user-images.githubusercontent.com/535300/52148680-48e01f00-261f-11e9-9d7e-136df62cd407.png)

velkymx:
[+ Add Rule]
if eval === true
assign to : USER or GROUP

velkymx:
NOTE that assignment form should include the Allow Reassignment and a Save button

PR #1329, opened 2019-02-06
Assign to User by Eval #1310 

danloa:
- Modeler Inspector was modified to add the new section for special assignments
- A new assignment rule function was created. This functions evaluates all the rules configured in a task. The first one that is true is used to make the assignment.
- processmaker-bpmn-moddle was modified to include a new attribute assignmentRules to store in the bpmn the rules to create an assignment by value (https://github.com/ProcessMaker/processmaker-bpmn-moddle/issues/10).
- package.json was updated to use the lates version of processmaker-bpmn-moddle


rcgongora:
this new Feature works as expected, but I a had a problem trying to understand how adding a new assignment rule by evaluation. When "Special Assignments Expression" is empty the assignment rule is not created, maybe if we add a validator that says: "This field is required to add the assginment rule"

danloa:
@rcgongora  The change has been implemented

PR #1330, opened 2019-02-06
Implementation of Screen Builder CSS

### LINKED ISSUE #1313: Screen Builder CSS
Track as part of the issue in the form builder: https://github.com/ProcessMaker/vue-form-builder/issues/46

- [x] Add the custom css property from vue-form-builder
- [x] Save the custom css configuration
- [x] Add the custom css property to the process runtime
- [ ] New release of vue-form-builder and update in package.json

caleeli:
Fixes #1313

*NOTE* remember to use the new version in package.json after vue-form-builder is deployed. Currently it's pointing to feature/55 branch until it's deployed

This PR includes:

- Added the custom css property from vue-form-builder.
- Add the use of custom css property in process runtime.
- Saved the custom css property.

image

image


rcgongora:
In the Custom CSS window There are no Cancel and Save buttons, How Can I revert the changes or cancel changes?

PMQABen:
You have to press the close button on the Custom CSS Modal, and then press the save icon on the far right, to save the Screen you're editing.

It looks like if a user opens up the CSS scope a bit to just plain old 'div' they could get changes that end up modifying the entire task form, but it only affects the form page itself.
image

tdondich:
Yeah, we need these rules to be absolutely scoped to only the form render, not outside.

PMQABen:
I grabbed the CSS from https://codepen.io/cb_rn/pen/OVXXmj if y'all want to take a better look at it.

nolanpro:
@PMQABen @caleeli I've updated the branch in vue-form-builder to use a library to parse and add the wrapper selectors. It has the added benefit of css syntax checking. For now, we're not allowing at-rules (like `@import`). I think this is OK seeing that we don't want to give them too much control but we allow certain ones (like `@media`) if the need is there.

nolanpro:
@PMQABen finally got this to deploy correctly. Have at it https://feature-1313.bpm4.qa.processmaker.net

PMQABen:
image

That'll do nicely. 👍

velkymx:
@caleeli let's try and do this ticket as part of this PR. I wonder if we can reduce dependencies while picking up the editor features and linter.

https://github.com/ProcessMaker/bpm/issues/1347

nolanpro:
@caleeli I updated monaco in this branch. Looks like its working fine. Let me know if you see anything. I was going to look at #1347 when I'm done with the webhooks issue but feel free to start it if you want

Issue #1328, opened 2019-02-04
User is unable to update own profile without "edit users" permission


PMQABen:
1. Log in as admin
2. Create a new user
3. Log out and back in as the new user
4. Go to the user's profile page `/profile/edit`
5. Change any part of the user's profile and press update

**Result:**
```
PUT http://bpm4.local.processmaker.com/api/1.0/users/23 403 (Forbidden)

[Vue warn]: Error in render: "TypeError: Cannot read property 'firstname' of undefined"
    vendor.js?id=03b5b59fa3a3fa0d7818:53823
    dispatchXhrRequest @ vendor.js?id=03b5b59fa3a3fa0d7818:188
    xhrAdapter @ vendor.js?id=03b5b59fa3a3fa0d7818:25
    dispatchRequest @ vendor.js?id=03b5b59fa3a3fa0d7818:593
    Promise.then (async)
    request @ vendor.js?id=03b5b59fa3a3fa0d7818:420
    Axios.(anonymous function) @ vendor.js?id=03b5b59fa3a3fa0d7818:440
    wrap @ vendor.js?id=03b5b59fa3a3fa0d7818:827
    onUpdate @ edit:407
    invoker @ vendor.js?id=03b5b59fa3a3fa0d7818:55257
    fn._withTask.fn._withTask @ vendor.js?id=03b5b59fa3a3fa0d7818:55056

TypeError: Cannot read property 'firstname' of undefined
    vendor.js?id=03b5b59fa3a3fa0d7818:54971 TypeError: Cannot read property 'firstname' of undefined
    at Proxy.eval (eval at createFunction (vendor.js?id=03b5b59fa3a3fa0d7818:63910), <anonymous>:3:545)
    at Vue._render (vendor.js?id=03b5b59fa3a3fa0d7818:57774)
    at Vue.updateComponent (vendor.js?id=03b5b59fa3a3fa0d7818:56018)
    at Watcher.get (vendor.js?id=03b5b59fa3a3fa0d7818:56372)
    at Watcher.run (vendor.js?id=03b5b59fa3a3fa0d7818:56449)
    at flushSchedulerQueue (vendor.js?id=03b5b59fa3a3fa0d7818:56211)
    at Array.<anonymous> (vendor.js?id=03b5b59fa3a3fa0d7818:55067)
    at flushCallbacks (vendor.js?id=03b5b59fa3a3fa0d7818:54988)

[...]

PR #1331, opened 2019-02-06
User is unable to update own profile without "edit users" permission #1328

danloa:
This PR solves the issue #1328

- User policy file added.
- Routes and controller modified to use the new policy.


PMQABen:
Fix works as expected.

Issue #1332, opened 2019-02-06
Server Error when adding task assignment without target


PMQABen:
1. Log in
2. Create a process
3. Add a task
4. Change the task assignment to be either `to user` or `to group`
*Note: Do not select an assigned user or group
5. Save the process

**Result:**
"Server Error" alert on the top of the page
```
/api/1.0/processes/2:1
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
﻿
vendor.js?id=03b5b59…:188
PUT http://bpm4.local.processmaker.com/api/1.0/processes/2 500 (Internal Server Error)
```
**Expected:**
An error message that tells the user to be sure they've chosen a user or group for all assigned tasks.

danloa:
@PMQABen  do you think that improving the message but using the current alert at the top is ok? Or it will be better to use a pop up to show the message?. Personally I'm biased to use the alert cause is more consistent with the rest of the application

PMQABen:
From a UI perspective, I would prefer helper text to appear in red (or some other eye-grabbing color), when a Task Assignment is changed to User or Group. However, I would be fine with making the error message more verbose and keeping the alert. 😃

PR #1336, opened 2019-02-08
Server Error when adding task assignment without target #1332

danloa:
This ticket solves #1332 

- The error message was improved to instruct to the user that he should select one specific user or group when using a group or user assignment.
- If the assignment is incorrect the server responds with status 422 instead of 500 as before.


rcgongora:
The error message is displayed when we try to save the process, and the error is 442 instead of 500. It works as expected

PR #1345, opened 2019-02-12
Set Cancel Screen

### LINKED ISSUE #1339: Set Canceled Screen
From the process configuration screen I want to be able to assign a screen to show in the event that the process is canceled.

milaendo:
closes #1339 

In the process config there is now another drop down which will only show "DISPLAY" screens as options for the cancel screen. So to test, first create a new screen with the type of display. Then open a processes config, select your new screen as the cancel screen. Start a process and cancel it. 

screen shot 2019-02-12 at 9 04 09 am
screen shot 2019-02-12 at 9 04 29 am


PMQABen:
Works as intended.

ProcessMaker/Http/Controllers/ProcessController.php
```diff
@@ -58,8 +58,8 @@ public function edit(Process $process)
             ->get()
             ->pluck('name', 'id')
             ->toArray();
-        
         $screens = Screen::orderBy('title')
+            ->where('type', 'DISPLAY')
             ->get()
```
velkymx: Do we need get if we are plucking?

ProcessMaker/Http/Controllers/RequestController.php
```diff
@@ -87,6 +87,11 @@ public function show(ProcessRequest $request, Media $mediaItems)
         $request->user;
         $request->summary = $request->summary();
         $request->summary_screen = $request->getSummaryScreenId();
+
+        if ($request->status === 'CANCELED' && $request->process->cancel_screen_id) {
+            $request->cancel_screen = $request->process->cancelScreen;
```
velkymx: This should set the summary screen ... not add a new one.

resources/views/requests/show.blade.php
```diff
@@ -277,7 +277,11 @@ class="nav-link">
                  * Get Screen summary
                  * */
                 screenSummary() {
-                    return this.request.summary_screen.config;
+                    if(this.request.status === 'CANCELED') {
+                        return this.request.cancel_screen.config;
```
velkymx: Set the summary screen, dont add a new one.

PR #1357, opened 2019-02-13
Webhooks for start events

### LINKED ISSUE #1342: Webhooks
I want to be able to start a process by posting to a specific URL. The URL should contain a token value. This token should be revocable. Add to start event inspector (webhook). Upon selection expose the URL w the token.

nolanpro:
Fixes #1342

- Allows the user to generate a publicly accessible URL that can be POST'ed to in order to start a new request. 
- Data can also be sent as part of the posted body and it will get saved in the request.
- Instead of "revoke" functionality, there is simply a disable link that does the same thing. When re-enabled, it will have a different url/token (the user is warned about this)
- Changing the node name will effectively revoke the hook but that's the same limitation we have for all resources associated by the node name, like assignments.
- Requests started from a webhook will show "Webhook" for Requested By in the UI


PMQABen:
So anyone can hit this POST URL, no filtering, no authorization required, etc?

nolanpro:
@PMQABen yup thats the intended behavior. No authorization (the token is the authorization) and anyone can start a process as long as the webhook is active. I'll fix some of the copy on the message

danloa:
If the request doesn't have an Accept: application/json header, an Html page is returned
request

PR #1360, opened 2019-02-14
updated all config/edit screens to have uniform buttons

milaendo:
all config and edit screens now have the same save/cancel buttons in secondary class. 
screen shot 2019-02-13 at 4 50 52 pm


PMQABen:
Looks good.

PR #1392, opened 2019-02-19
scroll bar added to request modal

### LINKED ISSUE #1391: New Requests off Screen
When starting a new Request, the "New Request" page displays. If there are lots of Categories (as there are now in develop-demo server), the Categories fall off the screen. There is no scroll bar to scroll down the screen to view the long list of Categories. To fix this, add a scroll bar.

milaendo:
closes #1391 

to test: 
create multiple categories and multiple processes inside of those categories.
open the request modal 
the modal itself now scrolls rather than the entire screen


PMQABen:
👍

PR #1401, opened 2019-02-20
API returning 10

milaendo:
In the modeler screens, users, and groups are no longer only returning 10. 
screen shot 2019-02-20 at 10 21 17 am
screen shot 2019-02-20 at 10 29 42 am


velkymx:
I'm still having the issues when selecting a script.

velkymx:
While adding to return a very large record set is an OK work around, we need to update the API to support unlimited returns. I'm going to reopen this issue to address.

PR #1404, opened 2019-02-20
screen validation

### LINKED ISSUE #1365: Screen Builder Validation
Even when validation errors are present, I can still save a screen.

milaendo:
closes #1365 

When saving a screen with validation errors, you will not longer be able to save it.


velkymx:
@milaendo can you add an alert that tells them way they cannot save. "This screen has validation errors." type danger

milaendo:
@velkymx Will do!

milaendo:
screen shot 2019-02-21 at 12 22 34 pm
@velkymx Message was updated

PR #1405, opened 2019-02-21
Update screen seed files

### LINKED ISSUE #1377: Update Seeders
Update the seeders to use updated data. The current seeder is for a VERY old process and needs to be replaced with a new one with new updated assets built with the latest tools.

nolanpro:
Fixes #1377 

Updated all screen seeds so they include the new inspector controls


velkymx:
This isn't what was discussed. I've updated the issue with more details.

PR #1416, opened 2019-02-22
UI changes for group membership config inside /admin/users

### LINKED ISSUE #1403: UI changes for group membership config inside /admin/users
1. Log into `/admin/users`
2. Edit a user
3. Go to the Groups tab
4. Click `+Add User to Group`

image
Buttons need to use the correct class formatting, the header text of the modal should be in black as opposed to blue.

----

image
Buttons need to use the correct class formatting. Message needs to be rewritten:
`Are you sure you want to remove the user from the group !!GROUPNAME!!?`

----

1. Log into `/admin/users`
2. Edit a user
3. Go to the Groups tab
4. Click `+Add User to Group`
5. Do not choose a group, press OK

**Result:** Nothing happens. No error message is thrown on screen or in the console, but the modal remains displayed.

----

image
Can we move the modal further down the page? It would prevent stuff like this from occurring:
image

----

One last potential change, why does the drop-down menu show groups the user is already in? If a user is in a group called "foobar", I probably should not see "foobar" as a selectable choice if I'm adding them to a new group.

marcoAntonioNina:
Fixes #1403 

- Add new api for avaible group and users in members
- multiselect enable filter remote with api.
- change modal in assigne new groups


PMQABen:
Drop-down menu only shows 10 groups at a time. If a company has more than 10 groups, they will not appear on the menu unless you search for them. And there's nothing in the drop-down menu that suggests a user can type to search for a group.

marcoAntonioNina:
yes, it only shows 10 elements, but the search is enabled its possible to enter text to filter elements.

PMQABen:
image
Where on that modal does it tell the user they can type something to search? 😄
When a user clicks the arrow to expand the menu, they're going to be looking at the list of group names, and likely will not notice the flashing cursor next to "Select option"

If the text was changed to something like `"Select group or type here to search groups"`, then the user would know they could type something.

PR #1444, opened 2019-02-27
Fix for showing the create category button when none exist

### LINKED ISSUE #1441: Process Category Required
There was a fix for this, but if you go to add a process, and no process category exists, then the modal should show a message that categories are required with a link to add a category.

screenshot 2019-02-27 10 23 59

nolanpro:
Fixes #1441


PMQABen:
image
If I remember correctly, we're supposed to capitalize the word Process for these kind of messages. 😄

resources/views/processes/index.blade.php
```diff
@@ -65,7 +65,7 @@
               <span aria-hidden="true">&times;</span>
             </button>
           </div>
-          <div class="modal-body" v-if="processCategories">
+          <div class="modal-body" v-if="hasProcessCategories">
```
velkymx: Cant we just check the length of the array in the v-if? I don't think we need all of this additional code...

nolanpro: processCategories is an object so it does not have a length property. The only way to check the length of an object is by counting its keys with `Object.keys(obj)`. The original implementation in https://github.com/ProcessMaker/bpm/pull/1106 was broken. It probably couldn't be tested because you can not delete the seeded category unless you delete the process from the db

ProcessMaker/Http/Controllers/ProcessController.php
```diff
@@ -36,7 +36,7 @@ public function index(Request $request)
         $status = $request->input('status');
         $processes = Process::all(); //what will be in the database = Model
         $processCategories = ProcessCategory::where('status', 'ACTIVE')->get();
-        $processCategoryArray = ['' => 'None'];
```
velkymx: We want this. It makes it so there is a default of none that validates that they must choose a category

resources/views/processes/index.blade.php
```diff
@@ -65,7 +65,8 @@
               <span aria-hidden="true">&times;</span>
             </button>
           </div>
-          <div class="modal-body" v-if="processCategories">
+          @if (!empty($processCategories))
```
velkymx: Change this to ensure that the count($processCategories) > 1)

ProcessMaker/Http/Controllers/ProcessController.php
nolanpro: Ok I put it back. I originally took it out because it is being overwritten by `[null => '']` on line 83 in the blade template.

PR #1446, opened 2019-02-27
Properties for New DatePicker

### LINKED ISSUE #1445: Properties for New DatePicker

marcoAntonioNina:
Fixes #1445 

- Add calendar format in file json.
- add variable in ProcessMaker. user.calendar_format.
- add overwrite css of calendar.


marcoAntonioNina:
The library that was added is not compatible with moment.js has its own syntax.

PR #1486, opened 2019-03-06
Translations

### LINKED ISSUE #1471: Update Translation File
Make sure the translation file has all of the keys we're using throughout the code base. Test by inserting an different language file (chinese maybe?) and check that the entire site translates properly. 

- [x] Add any missing translations to language file
- [ ] Add tests to confirm working
- [x] Note any places where there are not translations

milaendo:
Closes #1471 
There has been a new file created called en.json This is the only file that we need for both javascript and php translations. I have also gone through the app and added and missing translation tags. Listed in the original ticket are the places we will need to dig into a bit more to ensure they translated properly.  

What this PR does not include are tests for the translations. I believe this should be in the scope of the future ticket for the implementation of the translations. Please push back if this is not correct.


PMQABen:
https://feature-1471.bpm4.qa.processmaker.net/admin/groups/1/edit
image
 
image
image

PMQABen:
Unable to find anything else out of the ordinary. 👍

Issue #1483, opened 2019-03-06
Installer Tweaks


velkymx:
The installer is far to complex and need to be adjusted.

- [x] Remove email installer
- [x] Hide output of migrations
- [x] Remove CRON
- [x] Remove docker
- [x] Add dusk and faker to the prod requirements
- [x] Send messaging to Louis to update in spreadsheet
- [x] Always install group seeder
- [x] Add PHP Zip check to dependencies
- [x] Remove check for SOAP

velkymx:
Copy sent to Louis for revision.

PR #1487, opened 2019-03-06
Updated Installer

velkymx:
The installer has been updated to remove all of the setup for email, docker and cron to satisfy #1483

### DIFF
--- a/ProcessMaker/Console/Commands/Install.php
+++ b/ProcessMaker/Console/Commands/Install.php
@@ -167,10 +167,6 @@ class Install extends Command
 		//Create a symbolic link from "public/storage" to "storage/app/public"
         $this->call('storage:link');
 
-        // Install Docker executors
-        $this->info(__('Installing Docker executors...'));
-        $this->installDockerImages();
-
         // Restart services so they pick up the new settings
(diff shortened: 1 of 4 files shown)


ProcessMaker/Console/Commands/Install.php
```diff
@@ -153,36 +141,23 @@ public function handle()
...
-        
-        // Install Docker executors
-        $this->info(__('Installing Docker executors...'));        
-        $this->installDockerImages();
-        
         // Restart services so they pick up the new settings
-        $this->info(__('Restarting Services...'));
-        $this->info(
+        $this->info(__('Restarting services...'));
+        $this->info(e
```
ryancooley: Extra character "e" is causing some issues here
```suggestion
        $this->info(
```

ryancooley: Other than that change above this looks good and functions as intended.

tdondich: No services should be handled by the installer.  What's it trying to restart?

velkymx: I've removed it

PR #1490, opened 2019-03-07
Add error display in script editor.

### LINKED ISSUE #1481: Script Log Missing
When the docker container returns an error the error should be added to the output array of the execute console log. This was working. We need more details on the script.

Ex: 

Docker Not Found
Downloading Docker Image
Executing Script
Script Error: Output

caleeli:
Fixes #1481 

This PR adds error report when the script fails.

screen shot 2019-03-07 at 4 25 17 pm

image


PMQABen:
image
There are now 3 "interactive" button on the execute line, but only the one in blue works. This technically passes in a "it works as written" situation, but the UI needs polish.

caleeli:
Removed unused clickable buttons.

PR #1496, opened 2019-03-08
Script API User #1469

### LINKED ISSUE #1469: Script API User
The script configuration should allow an API user to be set for when the script runs and accesses our API via the SDK. If there isn't a user assigned, surface an alert on the scripts list to notify the users that there isn't a user that is configured to access the API. 

- [x] Code tests to confirm the failure and success as well as to confirm user assignment
- [x] Add config select for a single user having a token to be assigned to a script
- [x] Add warning when no user tokens exist
- [x] Save the API user to the script

danloa:
Fixes #1469 

- Run As select option added to edit and index screens
- Added tests that verify that when running a script it must have a user assigned 
- Alerts when "run as user" is empty or the selected user doesn't have an access token added.
- Scripts and Screen Versions tables modified to include the run_as_user_id field
- Many tests were modified because now the run_as_user_id is mandatory.

user


ProcessMaker/Http/Controllers/Api/ScriptController.php
```diff
@@ -202,6 +204,15 @@ public function store(Request $request)
         $request->validate(Script::rules());
         $script = new Script();
         $script->fill($request->input());
+
+        $userToRunAs = User::find($script->run_as_user_id);
+        if ($userToRunAs->tokens()->get()->count() === 0 && !$userToRunAs->is_administrator) {
+            return response(
+                ['message' => "The user doesn't have an access token",
+                    'errors' => ['run_as_user_id' => ["The user doesn't have an access token"]]],
```
velkymx: I shouldn't be able to choose a user if they don't have a token. If no users exist, you shouldn't be able to create a script and the modal should change in the same way the modal for adding processes changes when there are not process categories.

PR #1501, opened 2019-03-11
Vagrant provisioning: fixing Docker install

### LINKED ISSUE #1456: Docker install failing in vagrant after.sh - causes script tasks to break
Docker changed their package names and locations so fresh vagrant builds fail for bpm.

We need to change in after.sh:
```
sudo apt-get -y \
    -o Dpkg::Options::="--force-confdef" \
    -o pkg::Options::="--force-confold" \
install docker.io
```
To this:
```
curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
sudo add-apt-repository \
   "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
   $(lsb_release -cs) \
   stable"

sudo apt-get update
sudo apt-get -y install docker-ce docker-ce-cli containerd.io
```

Then vagrant destroy && vagrant up

ryancooley:
Fixes #1456. Docker now installs property in new Vagrant machines.

### DIFF
--- a/after.sh
+++ b/after.sh
@@ -9,10 +9,14 @@
 # which will be run after this script.
 
 # Install docker, which we use to run scripts
-sudo apt-get -y \
-    -o Dpkg::Options::="--force-confdef" \
-    -o pkg::Options::="--force-confold" \
-install docker.io
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+sudo add-apt-repository \
+   "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
+   $(lsb_release -cs) \
+   stable"
+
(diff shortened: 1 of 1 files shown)


after.sh
```diff
@@ -9,10 +9,14 @@
 # which will be run after this script.
 
 # Install docker, which we use to run scripts
-sudo apt-get -y \
-    -o Dpkg::Options::="--force-confdef" \
-    -o pkg::Options::="--force-confold" \
-install docker.io
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+sudo add-apt-repository \
```
velkymx: What happens with a CentOS system?

PR #1533, opened 2019-03-14
Create EnvironmentVariablesTest.php file

PMQABen:
Added a Dusk file to test adding/editing/deleting environment variables


PMQABen:
This keeps failing either on an npm critical vulnerability warning, or a Sauce Labs test that seems to work fine for other branches.

PR #1540, opened 2019-03-18
Adding a space to a process identifier will cause a saving to throw a 422 

### LINKED ISSUE #1307: Adding a space to a process identifier will cause a saving to throw a 422
_#1306 (potentially related)_
1. Create a Process
2. Change the Process Identifier to use a space (for example: `ProcessId` » `Process Id`)
3. Save the Process
4. Reload the page
5. Try saving again

**Result:**
```
The bpm definition is not valid

PUT http://bpm4.local.processmaker.com/api/1.0/processes/5 422 (Unprocessable Entity)

0: "Error 1824 (Line:3):Element '{http://www.omg.org/spec/BPMN/20100524/MODEL}process', attribute 'id': 'Process Id' is not a valid value of the atomic type 'xs:ID'."
1: "Error 1824 (Line:5):Element '{http://www.omg.org/spec/BPMN/20100524/DI}BPMNPlane', attribute 'bpmnElement': 'Process Id' is not a valid value of the atomic type 'xs:QName'."
length: 2
```
[process.mp4.zip](https://github.com/ProcessMaker/bpm/files/2818929/process.mp4.zip)

Even removing the space and saving, the user still runs into a 422.

danloa:
Resolves #1307

- The identifiers are stored as part of bpmn ids so it's not allowed to use spaces
- A validation to prevent storing of ids with empty spaces were added
- Tests to validate and generate the errors were added.


PMQABen:
The validation works, but it doesn't explain to the user what the problem is, just shows up with an alertbox message saying `"The bpm definition is not valid"`.

If you dig into the developer tools, the console has a slightly more informative error message: `"DOMDocument::schemaValidate(): Element '{http://www.omg.org/spec/BPMN/20100524/MODEL}process', attribute 'id': 'Process Id' is not a valid value of the atomic type 'xs:ID'."`

PR #1549, opened 2019-03-19
Add/remove current user in tasks.

### LINKED ISSUE #1542: Current user for the task (script or user) should be added to the json model and removed once completed
When executing a task, the user assigned to the task or currently performing the task should be added to the json data ONLY FOR THE DURATION OF THE TASK.  The property should be called _user

This allows the current screen or script to reference the currently used user that is executing.

Please note, proper attributes such as password should be omitted from the data model.

caleeli:
Resolves #1542 
This PR includes:

- Add _user to request data when an activity (human task or script) is activated.
- Removes _user from request data when an activity (human task or script) is completed.

image

image

### DIFF
--- a/ProcessMaker/Models/DataStore.php
+++ b/ProcessMaker/Models/DataStore.php
@@ -101,4 +101,17 @@ class DataStore implements DataStoreInterface
     {
         return $this->itemSubject;
     }
+
+    /**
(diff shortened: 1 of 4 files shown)


PMQABen:
Works as expected 👍

PR #1561, opened 2019-03-21
Verify Import

### LINKED ISSUE #1553: Verify Import
Importing a process throws a 500 error, but the process/category still imports successfully (scripts do not import). We should add a success page that shows all items that could be imported and the status of said import.

Something like:

[ check ] Process BPMN
[  x  ] Script Task

marcoAntonioNina:
resolves #1553 

- Add modal list of imported options.

image

image


marcoAntonioNina:
added class.
image

PR #1564, opened 2019-03-21
Create EnvironmentVariablesTest.php Artisan Dusk file 

PMQABen:
Daniel says the current CircleCI build process doesn't like it when we have capital letters in the branch name. Closed the previous pull request, renamed the branch to lower-case, making new request.


velkymx:
This is still having issues. I don't think it ever worked. Should not have been put up as a PR.

PR #1579, opened 2019-03-25
New User Default Values

### LINKED ISSUE #1567: New User Default Values
When adding a user, the drop-down menu has three choices: "blank", " active, and "inactive". We should change the blank to a placeholder text along the lines of "Select Status"

marcoAntonioNina:
Resolves #1567 

- Add placeholder in dropdown status in form create user


PMQABen:
👍

PR #1584, opened 2019-03-25
Add 3-legged auth to swagger ui

### LINKED ISSUE #727: Add OAuth login to swagger UI

nolanpro:
Fixes #727 
Populates the client_id and client_secret automatically

Kapture 2019-03-25 at 13 06 20


tdondich:
@nolanpro and @velkymx , is it possible to not show the client id and secret as fields in the authorize dialog? This should automatically be populated and not be able to be changed.

nolanpro:
Thats controlled by swaggerUI and [hiding it is not supported](https://github.com/swagger-api/swagger-ui/issues/4448#issuecomment-382563529), however, we can hide it with css. I'll add that.

PR #1585, opened 2019-03-25
Session Timeout Login

### LINKED ISSUE #1524: Session Timeout
- [x] Session background refresh
- [x] For components that emit messages, we need to add a listener that automatically resets the session
- [x] If a user is working in multiple tabs, all tabs must share the same session count down. 
- [x] Modal should show when a session is 95% expired
- [x] Modal progress bar should be dynamic with the session time
Screen Shot 2019-04-09 at 4 12 25 PM

milaendo:
closes #1524 

Screen Shot 2019-03-25 at 4 21 52 PM

to test:

In session.php set the 'lifetime' => 120, to 1.
load the page and wait one minute, you will then be prompted with a modal to log back in.


velkymx:
@milaendo @ryancooley I'm getting errors and it's not working. I did a fresh install (npm ci) and fresh db seed. The modal shows, but doesn't refresh my session.

Screenshot 2019-03-26 10 38 18

PR #1604, opened 2019-03-27
Import / Export Screens 

### LINKED ISSUE #1582: Import / Export Screens
I want to be able to export and import screens by themselves. 

- [x] Add export to the top right in the toolbar of the screen builder
- [x] Export a JSON object - name of the screen + .pm4
- [x] On the Screen List - add import button
- [x] Import button modal: Title: Import Screen, Body: Browse for File, Buttons: Cancel / Import
- [x] Include tests

marcoAntonioNina:
Resolves #1582 

- Add api export import screen.
- Add options export in screen builder and list screens.
- add import in list screens
- add test export and import screens


PMQABen:
If a screen cannot be imported, we don't tell the user why it can't be imported, there are no detailed errors shown in the console or network.

Other than that, import/export works fine.

PMQABen:
When exporting a screen, the user sees both the "saved successfully" banner notice, along with "the screen was exported" message. Is this intended behavior?

If a user does not have export permissions, they can still click the export button inside the Screen editor. It results in an error message like this:
image

If an admin enables export permissions for a user, and the user does not log out and back in, when clicking the export button they get the following error:
```
Not Authorized

Contact your administrator for more information
```

velkymx:
@marcoAntonioNina please include some unit tests

PR #1605, opened 2019-03-28
Browse Away Warning

### LINKED ISSUE #1595: Browse Away Warning
When I try to leave a screen builder or modeler editor I want a popup that halt the browser with a message, "Are you sure you want to leave this screen? Unsaved changes will be lost!" Add a button for confirmation of nav away, and a second button to save the screen.

marcoAntonioNina:
Resolves #1595 

Add alert when the user wants to leave the page.

pages Screen builder and modeler.

note. 
The onbeforeunload event.

The default message that appears in the confirmation box, is different in different browsers. However, the standard message is something like "Are you sure you want to leave this page?". This message cannot be removed.

### DIFF
--- a/resources/views/processes/modeler/index.blade.php
+++ b/resources/views/processes/modeler/index.blade.php
@@ -45,9 +45,19 @@ ol.breadcrumb {
 
 @section('js')
   <script>
-    window.onbeforeunload = function () {
-      return 'Are you sure you want to leave?';
-    }
+    new Vue({
+      created() {
+        window.addEventListener('beforeunload', this.handler)
+      },
+      methods: {
+        handler: function handler(event) {
+          let confirmationMessage = __('Are you sure you want to leave?');
+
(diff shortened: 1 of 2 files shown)


resources/views/processes/modeler/index.blade.php
```diff
@@ -45,6 +45,9 @@
 
 @section('js')
   <script>
+    window.onbeforeunload = function () {
+      return 'Are you sure you want to leave?';
```
velkymx: This string needs to translate

velkymx: Let's handle this with Vue. Something like...

```
new Vue({
  created() {
    document.addEventListener('beforeunload', this.handler)
  },
  methods: {
    handler: function handler(event) {

    }
  }
])
```

velkymx:
@marcoAntonioNina where is the test for this?

PR #1610, opened 2019-03-29
Close failed script task when Resquest is closed.

### LINKED ISSUE #1572: Error State
If a task fails in an error state, this will cause the whole request to fail in error. Allow the administrator to "Complete" the request at that point. Make sure the error is referenced in the log/notes and who forced it closed.

caleeli:
Resolves #1572 

This PR includes:

- Update Nayra
- Close a process in failing state
- Close scripts when process is closed.


PMQABen:
The "complete" button does not go away after you have manually completed the request. The request also does not go into the "completed" filter.

PR #1628, opened 2019-04-08
Cant Save Profile without Password

### LINKED ISSUE #1612: Cant Save Profile without Password
I should be able to save my profile without changing my password. It currently requires I change my password even when I don't want to. Validation should happen only if the first password box contains something.

Screenshot 2019-04-01 09 43 30

marcoAntonioNina:
Resolves #1612 

- Remove password of data when edit information user if field is empty.

### DIFF
--- a/resources/views/admin/users/edit.blade.php
+++ b/resources/views/admin/users/edit.blade.php
@@ -599,9 +599,11 @@
           },
           validatePassword() {
             if (!this.formData.password && !this.formData.confpassword) {
+              delete this.formData.password;
               return true;
             }
             if (this.formData.password.trim() === '' && this.formData.confpassword.trim() === '') {
+              delete this.formData.password;
               return true
             }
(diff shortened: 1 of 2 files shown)


PMQABen:
I tested the problem on what we have up on "develop.bpm4.qa", and I could not reproduce it. I created a separate user, logged in as them, and every time I saved a change to any part of that user's profile, it never prompted me to enter the password. The only time I was prompted to enter the password is if I intentionally went to update the new password field.

I'm not certain this is still an issue.

velkymx:
@PMQABen I had it happen to me. Could be autocomplete causing the issue?

PMQABen:
Yeah it looks like autocomplete may be at fault. I logged in with Firefox, and it only happened after I told the browser to save my login information.

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

PR #1630, opened 2019-04-08
over ride login redirect

### LINKED ISSUE #1615: Login Redirect
When a user signs in, the current behavior is to redirect you back to the page that timed out (like a 404 on the favicon), we had been hard coding the redirect and that need to be put back. 

All logins should go to /requests

milaendo:
closes #1615 

new method in login controller will now override the redirect path even if the page you were previously on was a 404.


ProcessMaker/Http/Controllers/HomeController.php
```diff
@@ -10,12 +10,6 @@ class HomeController extends Controller
 {
     public function index(Request $request)
     {
-
-      if (Auth::check()) {
```
velkymx: This should still exist. It checks if the user is logged in.

milaendo: The auth in the route is already checking so I thought it might be redundant but I can put it back

PR #1637, opened 2019-04-11
update database name

milaendo:
readme updated to point to the right database


milaendo:
@velkymx the workflow.sql file was removed. I confirmed with @nolanpro it was merged in on accident and all other references has been updated.

PR #1649, opened 2019-04-12
Remove htmlspecialchars since that happens automatically on output

### LINKED ISSUE #1648: Screen Names with Ampersand Showing Code
Screenshot 2019-04-12 11 04 16

Furthermore, if you change the Screen name from one with said ampersand to "and" you get double coded ampersands. Ryan says, "Very strange behavior".

Screenshot 2019-04-12 11 05 09

nolanpro:
Fixes #1648 

It was double encoding

@PMQABen need to hammer this again with injection, etc.

### DIFF
--- a/ProcessMaker/Http/Middleware/SanitizeInput.php
+++ b/ProcessMaker/Http/Middleware/SanitizeInput.php
@@ -53,10 +53,6 @@ class SanitizeInput extends TransformsRequest
             // Remove most injectable code
             $value = strip_tags($value);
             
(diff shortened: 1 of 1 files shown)


PMQABen:
The following will display as empty when used
```
<foo val=“bar” />
<foo val=“bar” />
<foo val=”bar“ />
<foo val=`bar' />
```
If anything gets wrapped in parentheses like `<hello>`, the string gets removed.

PR #1651, opened 2019-04-15
Add a Play / Pause action button the the process list

### LINKED ISSUE #1646: Play / Pause
Add a Play / Pause action button the the process list to allow for pausing of process that have timer start events. Remove the check box from config. Hide if there isn't a timer start event.

caleeli:
Resolves #1646 
This PR includes:
- Process::hasTimerStartEvents() which returns true|false if the process has Timer Start Events.
- Show|Hide configuration option in Process List

image

image

### DIFF
--- a/ProcessMaker/Models/Process.php
+++ b/ProcessMaker/Models/Process.php
@@ -728,7 +728,7 @@ class Process extends Model implements HasMedia
      *
      * @return boolean
      */
-    public function hasStartEvents()
+    public function hasTimerStartEvents()
     {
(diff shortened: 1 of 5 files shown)


milaendo:
This ticket also requires this option to be moved to the process as an action button 
Screen Shot 2019-04-15 at 9 19 06 AM

caleeli:
@milaendo Oh I will add it, thanks for the observation

caleeli:
updated

PR #1659, opened 2019-04-16
Can't hide File Upload / Download

### LINKED ISSUE #1591: Can't hide File Upload / Download
The File Download control in Screens Builder cannot be referenced in "Visibility Rules" expressions in other controls.

marcoAntonioNina:
Resolves #1591 

Add message by default when the component download file has no items to display.

image

image

image


marcoAntonioNina:
The validation rules are executed. only for this element can not be shown when there is no data, so text has been added.

PMQABen:
👍

PR #1664, opened 2019-04-16
Code that allows a package to add cutom screen builder components

danloa:
To register new components the packages must implement the following rule
In the root of the package folder a screen-builder-components.js file with the code to push the package's controls and  configurations


ProcessMaker/Managers/ScreenBuilderManager.php
```diff
@@ -1,16 +1,27 @@
 <?php
 namespace ProcessMaker\Managers;
 
+use Illuminate\Support\Facades\Storage;
+
 class ScreenBuilderManager
 {
     private $javascriptRegistry;
+    const SCREEN_BUILDER_EXTENSIONS_FILE = 'screen-builder-components.js';
```
velkymx: Please do not use constants.

PR #1669, opened 2019-04-17
Enable Rich Text in Screen Builder

### LINKED ISSUE #1624: Enable Rich Text in Screen Builder
Show the rich text icon and allow for usage in the Spark Screen Builder AND Form Builder

caleeli:
Resolves #1624 

image


PMQABen:
👍
Is there a reason we don't have this option for a Display-type Screen?

velkymx:
I agree with @PMQABen - why isn't this in the screen type display?

velkymx:
@caleeli add this control to the "DISPLAY" screen type

PR #1678, opened 2019-04-18
fix sticky breadcrumbs

milaendo:
breadcrumbs


velkymx:
That gif is wrong. The editor shouldn't scroll like that. It should fill the screen and the columns should scroll...

milaendo:
@velkymx would you like that to be fixed in this ticket or is that a screen builder thing?

milaendo:
The scrolling problem with screen builder will be solved in the new design of screen builder.

PR #1731, opened 2019-04-22
first and last name are now required

### LINKED ISSUE #1702: User's Name Required
When editing a user profile, it's possible to delete the First Name and Last Name field values and then save the profile. However, the edit profile screen is not dismissed (it just sits there).

To this fix, make the First Name and Last Name field values required so that a user's first name and last name are always there whenever someone mouse-hovers over the user avatar in lists.

Please tag Louis Criscuolo when this bug is fixed so I can document that First Name and Last Name fields are required.

milaendo:
closes #1702

Screen Shot 2019-04-22 at 2 18 01 PM


milaendo:
@lcriscuolo

lcriscuolo:
Thank you, @milaendo ! I very much appreciate you letting me know. I've revised documentation.

Could you please confirm if this also applies to a user's profile setting? I discovered this issue from editing a user profile: there might be a first and last name that an Admin enters for a new user, but that user could delete the First Name and Last Name values and the save the profile. This caused the default avatar to not display any user initials (it was just a yellow circle).

milaendo:
@lcriscuolo Thanks for pointing that out. I have added a new commit and can confirm that is required in both the user profile edit, and the admin create/edit user 
Screen Shot 2019-04-23 at 11 45 26 AM
Screen Shot 2019-04-23 at 11 46 20 AM

lcriscuolo:
Thank you, @milaendo . Might I suggest a couple suggestions in the "required" tooltips you added to the First Name and Last Name" fields? Please change "firstname" to "First Name" for the First Name field, and change "lastname" to "Last Name" in the Last Name field?

But I can adjust documentation to state that these fields are required. Thank you!

CC: @velkymx

milaendo:
@velkymx @lcriscuolo 
Screen Shot 2019-04-23 at 12 58 38 PM

lcriscuolo:
That looks awesome, @milaendo . Thank you so much!

PR #1744, opened 2019-04-23
Add info for loading, no results, and errors to vuetable

### LINKED ISSUE #1725: Vue-Table Loading Data
When vue-table is being loaded, we should show that the data is being loaded in the vue-table box. Once the table is loaded it should change to the grid. Same thing with empty data returned, should not show the grid, but a message that states no results were found.

nolanpro:
Fixes #1725

Couldn't find the exact icons so feel free to replace them. Here's the implementation for requests and tasks listings. You can test the error with `return abort(500)` in the controller.

Once this is good we can add it to anywhere we use vuetable2, or any other palace we need to wait for XHR to finish.

Kapture 2019-04-23 at 13 31 40


nolanpro:
@velkymx this has been updated. Let me know if you want to add this to everywhere we load a vuetable. Currently, it's only on Request listing and Task listing

Kapture 2019-05-07 at 17 20 50

PR #1762, opened 2019-04-29
Support display and edit form components added by packages

### LINKED ISSUE #1753: Support display and edit form components added by packages
Currently Screen Builder can just add components to edit screens (type FORM). Packages should be able to add FORM and DISPLAY components in the builder and renderer.

danloa:
Resolves #1753

- Every package that  registers new components in the screen builder, should publish the following files:
screen-builder-display-components.js To be used with the screen type DISPLAY
screen-builder-form-components.js To be user with the screen type FORM
- Changes to send the ScreenBuilderStarted when using the screen builder or when  the screen is rendered were implemented


ProcessMaker/Managers/ScreenBuilderManager.php
```diff
@@ -45,4 +39,27 @@ public function getScripts()
     {
         return $this->javascriptRegistry;
     }
-}
\ No newline at end of file
+
+    public function addPackageScripts($type)
+    {
+        // Depending of the form type we load the correct file with the components to display
+        switch ($type) {
```
velkymx: This logic could be clearer. We're adding a table for screen_types, can we use that instead?

danloa: The code was simplified with the convention that the files with the extension components use the screen type (form, display, etc.) in the file name.

PR #1765, opened 2019-04-30
Script editor redesign

### LINKED ISSUE #1750: Refresh Script Editor Design
Update to match the new screen builder.

image

tuantruong625:
- Use vue-bootstrap / bootstrap components whenever possible 

- Validation drawer will be implemented once linting is ready.

Fixes #1750


ssshake:
1) Needs some styling improvements. Because the list on the left needs to scroll excluding the debugger, (debugger acts as header) that broke the styling a bit. I'm not sure what the bootstrap was is to add the border back, dont want to do it manually for fear of breaking theming.

2) The scroll is overflowing itself, check the bottom of the scroll back for example

3) Output should auto expand if there is a result but it should not break manual expanding on click. Binding visible to preview.output it not good enough.

image

chipit24:
@velkymx I've updated the PR. Scrolling has been adjusted to work as expected, and the Output panel now auto-expands.

PR #1822, opened 2019-05-14
Fix FormText import

chipit24:
The PR fixes the `unknown custom element <FormText>` error in Spark. The Text element will now render in the screen builder preview, and data interpolation works as expected.


resources/js/processes/screen-builder/typeForm.js
```diff
@@ -3,10 +3,10 @@ import globalProperties from "@processmaker/spark-screen-builder/src/global-prop
...
 import FileUpload from "./components/form/file-upload";
 
-import {
-    FormText,
-} from "@processmaker/spark-screen-builder";
-import initialControls from "@processmaker/spark-screen-builder/src/form-builder-controls";
+import {renderer} from "@processmaker/spark-screen-builder";
+import {FormBuilderControls as initialControls} from "@processmaker/spark-screen-builder";
+
+const {FormText} = renderer;
```
tdondich: This line totally is invalid. FormText is not the renderer. It should be const {FormText} = renderer.FormText;

You can see this at https://github.com/ProcessMaker/spark-screen-builder/blob/develop/src/components/index.js#L24

ssshake: @tdondich I'm out of the loop here because it looks like this has been merged. But this is destructuring so this reads fine to me.

```const {FormText} = renderer```

is equivalent to 

```const FormText = renderer.FormText```

This maybe reads a little weird when you're only doing one assignment, but is really handy when doing multiple assignments.

example: 

```const {
    FormButton,
    FormMultiColumn,
    FormRecordList,
    FormText
} = renderer```

ssshake: ...I see that the merge was reverted due to an error referencing FormText so I'm probably missing the point. Will have to go over it with Rob tomorrow

PR #1833, opened 2019-05-16
Fixes database connection issue if using other hostname than localhost

### LINKED ISSUE #1832: Config uses incorrect env for database
When running php artisan spark:install, setup runs as expected. However after booting project up no database connection can be made. Issue resides in config/database.php where ENV DB_HOST is used instead of DB_HOSTNAME as is set up by spark:install.

alexatipscreener:
Fixes #1832

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -29,7 +29,7 @@ return [
     'connections' => [
         'spark' => [
             'driver' => 'mysql',
-            'host' => env('DB_HOST', '127.0.0.1'),
+            'host' => env('DB_HOSTNAME', '127.0.0.1'),
             'port' => env('DB_PORT', '3306'),
             'database' => env('DB_DATABASE', 'spark'),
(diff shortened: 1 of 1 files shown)


alexatipscreener:
If it's supposed to use DB_HOST and not DB_HOSTNAME as the ENV key variable, shouldn't the setup scripts also use the same ENV key variable? 
[Install command uses DB_HOSTNAME](https://github.com/ProcessMaker/spark/blob/771d57af65880f217292d106169161e2ffb90ec8/ProcessMaker/Console/Commands/Install.php#L192) while [database config uses DB_HOST](https://github.com/ProcessMaker/spark/blob/771d57af65880f217292d106169161e2ffb90ec8/config/database.php#L32)

velkymx:
@alexatipscreener I see what you're saying. Good catch!

PR #1846, opened 2019-05-17
Do not strip html tags from rich text content

### LINKED ISSUE #1843: HTML Tags Stripped
image

image

caleeli:
Resolves #1843 

image


caleeli:
I also verified expressions with "<" and ">" in calculated fields and css

image

image

velkymx:
@caleeli I'm going to merge this so HTML works, but we need to blacklist those form elements to satisfy the issue.

PR #1852, opened 2019-05-17
Fix display type form rendering

chipit24:
Fixes https://github.com/ProcessMaker/spark-screen-builder/issues/183, fixes https://github.com/ProcessMaker/spark-screen-builder/issues/187., fixes https://github.com/ProcessMaker/spark-screen-builder/issues/188.

Video of working form:
https://www.dropbox.com/s/ttorwt7je3t3hx6/spark_display_type_screen_working.mov?dl=0

Video of record list working:
https://www.dropbox.com/s/7bro86pr0ajrpri/record_list_working.mov?dl=0

**Note:** The display type screen now uses the same multi-column component as the form type screen. Apart from actually rendering and working now, the form control has a consisten name: `Table`, instead of `Multi Column`.


resources/js/processes/screen-builder/typeForm.js
```diff
@@ -4,12 +4,11 @@ import FileDownload from "./components/file-download";
...
 import initialControls from "@processmaker/spark-screen-builder/src/form-builder-controls";
 import VueFormElements from "@processmaker/vue-form-elements";
+import {renderer} from "@processmaker/spark-screen-builder";
 
 Vue.use(VueFormElements);
 
-import { renderer } from "@processmaker/spark-screen-builder";
-
-let FormText = renderer.FormText;
+const {FormText} = renderer;
```
ssshake: This was mis-identified in the past PR as being a problem. This is written correctly. Calling this out pre-emptively to prevent holding things up.

resources/js/processes/screen-builder/typeDisplay.js
```diff
@@ -1,10 +1,13 @@
...
 
-import {
+const {
     FormMultiColumn,
     FormText,
     FormRecordList
-} from "@processmaker/spark-screen-builder";
+} = renderer;
+
+const TableControl = FormBuilderControls.find(control => control.editorComponent === FormMultiColumn)
```
ssshake: The cause of this bug is that the 4 elements using the Display builder were manually defined in spark as a way to limit display builder to just these 4 elements. Because of this, as properties of these elements in screen-builder changed over time that meant these manual ones in spark got stale and broke things.

Rob removed the manual table/multi-form and replaced it with the one in screen-builder.

I believe the right move going forward is to do this for all 4 elements. I would like to discuss this as a group to make sure we have the right assumptions and then split that out into a new ticket.

ssshake:
Something is wrong with tests in develop and I don't know why because the latest circle ci tests passed for develop.

But in circle ci 6 tests are failing for this branch. Locally 4 tests are failing (less the two login tests failing in CI). 

I pulled down develop locally and the same 4 tests are failing.

velkymx:
@ssshake let me know about this issue. I'd like to get it merged.

PR #1886, opened 2019-05-27
Don’t trigger leave warning when there are no unsaved changes

chipit24:
Fixes https://github.com/ProcessMaker/spark-screen-builder/issues/209.

**How to test:**
1. Navigate to an existing screen or create a new one.
2. Make some changes and press the save button (ensure it successfully saves).
3. Navigate away from the page (pressing the refresh button is sufficient); you should **not** see the leave page warning.
4. Go back to the screen and make some changes, but this time, **do not** save them!
5. Navigate away from the page; you **should** see the leave page warning now.

### DIFF
--- a/resources/js/processes/screen-builder/screen.vue
+++ b/resources/js/processes/screen-builder/screen.vue
@@ -293,6 +293,8 @@ import Validator from "validatorjs";
       updateConfig(newConfig) {
         this.config = newConfig
         this.refreshSession();
+        ProcessMaker.EventBus.$emit("new-changes");
+
       },
       updatePreview(data) {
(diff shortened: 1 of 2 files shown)


resources/js/processes/screen-builder/screen.vue
```diff
@@ -293,6 +293,8 @@ import Validator from "validatorjs";
       updateConfig(newConfig) {
         this.config = newConfig
         this.refreshSession();
+        ProcessMaker.EventBus.$emit("new-changes");
+
```
tuantruong625: Extra line break

PR #1894, opened 2019-05-29
500 Modify Process

### LINKED ISSUE #1859: 500 Modify Process
Modify Document Process in BETA5 QA, if you try and start it, we get a 500. The process is probably not properly defined, but we should probably show something if there's a 500. Should show a friendly errors screen or alert with the exact issue so the user can fix it.

danloa:
Resolves #1859 

- The process controller has been modified to return more meaningful messages
- When some error happens at the moment of start a process an error message is displayed below the process name

image


velkymx:
@danloa https://github.com/ProcessMaker/spark/issues/1859#issuecomment-496967621

PR #1918, opened 2019-05-30
All Requests Permission Broken

### LINKED ISSUE #1906: All Requests Permission Broken
When creating a user, even after choosing to assign all permissions to that user, they are still unable to see the "All Requests" section of the requests page, nor are they able to see any tasks other than what have been assigned to them.

The only way it works is if the user is changed into a "Super Admin", but that's probably not a good idea.

milaendo:
closes #1906 

fixed the auth check so it now respects the "view all requests" permission.

to test: create a new user and give them "view all requests" permissions, login as that user and see all of the beautiful requests

### DIFF
--- a/resources/views/layouts/minimal.blade.php
+++ b/resources/views/layouts/minimal.blade.php
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="shortcut icon" type="image/x-icon" href="https://pbs.twimg.com/profile_images/521685439820742657/8B2oQKmP_400x400.jpeg" />
     <meta name="csrf-token" content="{{ csrf_token() }}">
-    <title>@yield('title',__('Welcome')) - ProcessMaker')</title>
+    <title>@yield('title',__('Welcome - ProcessMaker'))</title>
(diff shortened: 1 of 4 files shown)


resources/views/layouts/minimal.blade.php
```diff
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="shortcut icon" type="image/x-icon" href="https://pbs.twimg.com/profile_images/521685439820742657/8B2oQKmP_400x400.jpeg" />
     <meta name="csrf-token" content="{{ csrf_token() }}">
-    <title>@yield('title',__('Welcome')) - ProcessMaker')</title>
```
velkymx: Keep this

milaendo: its still there, just fixed the extra ')'. I can fix it is a different ticket if you would like!

velkymx: - ProcessMaker should always exist untranslated.

milaendo: gotcha!

PR #1968, opened 2019-06-07
Switch to PMQL queries for default request and task listings

### LINKED ISSUE #1960: Switch to PMQL queries for default request and task listings

milaendo:
closes #1960 

removed old code added in:  https://github.com/ProcessMaker/spark/pull/1829/files 
uses PMQL to only show in progress requests


resources/js/requests/components/RequestsListing.vue
```diff
@@ -113,6 +112,32 @@ export default {
       ]
     };
   },
+  beforeCreate() {	
+    //switch through all status
+    switch (Processmaker.status) {	
+      // if there is no status, meaning its on my requests, We should only show the in progress status
+      case "":
```
velkymx: Is this an empty case or the default case?

velkymx: Can we dry up this functionality?

milaendo: this is there is an empty case

PR #1981, opened 2019-06-12
False Value in Dropdowns

### LINKED ISSUE #1964: False Value in Dropdowns
image

danloa:
Resolves #1964 

- When a user is created his locale information is read from .env file or a default value is assigned.
- When editing a user if its locale is not configured, default values are selected.


resources/views/admin/users/edit.blade.php
```diff
@@ -190,17 +190,25 @@ class="mb-2"
                                             <div class="form-group col">
                                                 {!!Form::label('timezone', __('Time zone'));!!}
                                                 <b-form-select id="timezone" v-model="formData.timezone" class="form-control" :options="timezones">
-                                                  <template slot="first">
-                                                    <option :value="null" disabled>@{{appTimezone}}</option>
-                                                  </template>
+{{--                                                  <template slot="first">--}}
```
velkymx: Delete this... don't comment out

ProcessMaker/Http/Controllers/Api/UserController.php
```diff
@@ -121,16 +121,17 @@ public function store(Request $request)
...
         $fields = $request->json()->all();
-        if (isset($fields['password'])) {
-            $fields['password'] = Hash::make($fields['password']);
-        }
 
-        if (!isset($fields['timezone'])) {
-            $fields['timezone'] = env('APP_TIMEZONE');
-        }
+        // Default values if the users doesn't have set his locale configuration
+        $appTimezone = getenv('APP_TIMEZONE') !== false ? getenv('APP_TIMEZONE') : 'America/Los_Angeles';
```
velkymx: This should be part of the new user creation, in the database defaults.

ProcessMaker/Models/User.php
```diff
@@ -148,6 +148,22 @@ public static function rules($existing = null)
         'permissions',
     ];
 
+    protected static function boot()
+    {
+        parent::boot();
+
+        // Setting defaults values for a new user
+        static::creating(function ($model) {
+            $appTimezone = getenv('APP_TIMEZONE') !== false ? getenv('APP_TIMEZONE') : 'America/Los_Angeles';
```
velkymx: No. These default values should be set in the DATABASE migration (as part of the database default values)

PR #2014, opened 2019-06-16
Fix missing gateway tokens

### LINKED ISSUE #2013: Missing Gateway tokens
The following process is not running:

image

caleeli:
Resolves #2013 

This PR includes:

- Update Nayra version
- Save and update gateway tokens

image


PMQABen:
👍 Was able to create a Process like the above, and have it complete all tasks successfully.

PR #2028, opened 2019-06-19
Update file-upload component

chipit24:
Fixes https://github.com/ProcessMaker/screen-builder/issues/224.

Video of fix: https://www.dropbox.com/s/yrtl27hkak3xarg/file-upload_working.mov?dl=0

The PR fixes the `file-upload` component:
* Updates wording so user knows it a single file upload.
* No longer throws a console error when testing in the screen builder.
* Emits the file name instead of an ID.


PMQABen:
Those console errors still appear if you have the image variable name as `hmmyes.000.gif` or whatever `*.xxx.xxx` type extension.

chipit24:
@PMQABen Yup, I brought that problem up a couple weeks ago in a comment: https://github.com/ProcessMaker/screen-builder/issues/224#issuecomment-499235470. But the focus of that issue was shifted to getting the file-upload to work how @velkymx mentioned; the filename is treated as the form input value.

I'll open a separate issue for the naming tomorrow to keep the discussion more focused, but this PR will not deal with that.

PMQABen:
Ah, in that case this all checks out fine, works as expected. 👍

chipit24:
@tuantruong625 The issue with the masks exists with all form builder components; the mask's height is set to 50% and this causes it to overflow the thing it's suppose to over, or not cover it completely. This will have to be fixed in `screen-builder`, and I've created an issue for it there: https://github.com/ProcessMaker/screen-builder/issues/254.

PR #2037, opened 2019-06-24
Remove fs package

### LINKED ISSUE #2027: security holding package
Hi,

it seems you have accidentally added `fs` but this is not a package on npmjs, `fs` is an internal NodeJS module.

https://github.com/ProcessMaker/processmaker/blob/develop/package.json#L63

See https://nodejs.org/api/fs.html

You might want to remove it using `npm uninstall fs`.

caleeli:
Resolves #2027 

This PR removes the fs package because it is internal.


ZLTM:
Fix is working

PR #2048, opened 2019-06-27
Fix modeler container padding

chipit24:
This PR is for fixing https://github.com/ProcessMaker/modeler/issues/453 in `processmaker`. The associated PR in `modeler` is https://github.com/ProcessMaker/modeler/pull/454.

### DIFF
--- a/resources/js/processes/modeler/components/ModelerApp.vue
+++ b/resources/js/processes/modeler/components/ModelerApp.vue
@@ -14,7 +14,7 @@
         </div>
       </b-card-header>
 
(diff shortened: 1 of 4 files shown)


chipit24:
Updated to move the save button to the breadcrumbs bar:

Screen Shot 2019-06-28 at 11 53 48 AM

velkymx:
@chipit24 I've reverted this change out. The changes to the index template broke all of the other pages. Please revise and test to confirm other pages work as expected. A better solution would be to keeped the CSS scoped on that page to gain more control.

PR #2080, opened 2019-07-08
Fix query to get ACTIVE Intermediate Timer Events

### LINKED ISSUE #2079: Intermediate timer event runs only once in a loop process
The Intermediate timer event runs only once in the following process:

image

When the user back to the first task the second instance of the timer event never triggers.

[Loop_IntermediateTimerEvent.json.txt](https://github.com/ProcessMaker/processmaker/files/3369370/Loop_IntermediateTimerEvent.json.txt)

caleeli:
Resolves #2079 

This PR fixes the timer event for a loop process:

image


nolanpro:
I also tried this out and it looks good
image
image
image

PR #2121, opened 2019-07-15
New Request Modal Sorting

### LINKED ISSUE #2074: New Request Modal Sorting
Presently it seems Process Categories display in the "New Request" screen in the order that they were created. If an organization has many Categories, this will become unwieldy for requesters to locate a Process in a Category from which to start a Request. Granted a requester can search for a Process, but that user must know that Process already exists to know to search for it.

These Categories should be alphabetized. Furthermore, all Processes within each Category should be alphabetized. This will make it easier for Requesters to discover Processes when they are new users.

danloa:
Resolves #2074 

- The endpoint start_processes now supports multi-column sorting
- The call to get the process and categories list now requests the information ordered by category name and process name.


ProcessMaker/Http/Controllers/Api/ProcessController.php
```diff
@@ -451,18 +451,25 @@ private function saveTaskNotifications($process, $request)
     public function startProcesses(Request $request)
     {
         $where = $this->getRequestFilterBy($request, ['processes.name', 'processes.description', 'category.name']);
-        $orderBy = $this->getRequestSortBy($request, 'name');
+        $orderColumns = preg_split('/\,/', $request->input('order_by', 'name'));
```
caleeli: Why are you using regular expression of one only character instead of implode? If we are not going to accept spaces (for example) an `implode` could be more performant

danloa:
@nolanpro I added a test in ProcessTest.php to verify that the processes to start are ordered by process category and process name.

PR #2122, opened 2019-07-15
Modal Process Category Status

### LINKED ISSUE #2073: Modal Process Category Status
Add the "Status" drop-down to the "Create Category" screen and be done with it? The "Edit Category" screen is fine as-is, but don't require the user to see two screens to set one setting for a new asset.

marcoAntonioNina:
resolves #2073 

- Add dropdown status in create modal category.
- After create category refresh list categories


nolanpro:
LGTM

PR #2123, opened 2019-07-15
Fix request file upload permissions

### LINKED ISSUE #2084: Assigned User File Download Issue
When a user opens a task assigned to them, if the form has a File Download control but the user is not an admin, the Screen will show an Authorization error, and the file will be inaccessible. If the user has admin privileges, he'd see no error and be able to download the file just fine. This is unacceptable, since if the user has a task assigned, they should naturally be able to fetch download files part of the form. 

Replicate:
- Create a screen with a File Upload control
- Create a screen with a File Download control, same variable
- Add screens to two serial tasks, each assigned to different (non admin) users.
- Run the request, add a file as user one. Submit.
- Upon opening the assigned task, user two will encounter this (notice the error above, and the file control):

caleeli:
Resolves #2084 

This PR includes:

- Test to verify user permission
- Fix request file upload permissions to include the assigned user

image

image

[upload_file.json.txt](https://github.com/ProcessMaker/processmaker/files/3393099/upload_file.json.txt)

### DIFF
--- a/ProcessMaker/Policies/ProcessRequestPolicy.php
+++ b/ProcessMaker/Policies/ProcessRequestPolicy.php
@@ -92,4 +92,20 @@ class ProcessRequestPolicy
         
         return $user->can('editData', $request->process);
     }
+
+    /**
+     * User has access if participates in the request.
+     *
+     * @param  \ProcessMaker\Models\User  $user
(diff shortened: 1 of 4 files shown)


nolanpro:
See question here https://github.com/ProcessMaker/processmaker/issues/2084#issuecomment-511578175

PR #2126, opened 2019-07-15
Add support for password grant oauth flow

### LINKED ISSUE #2081: Auth Clients Grant Secret
Show the password grant client secret in the Auth Clients screen in administration

This needs to be visible so that people can use the password grant type effectively.

nolanpro:
Fixes #2081 

image


resources/views/admin/auth-clients/index.blade.php
```diff
@@ -53,6 +52,13 @@
 
         <div class="px-3 page-content">
             <div class="row">
+                <div class="col" align="left">
```
caleeli: Why are showing above the table only the first client with password grant enabled?
@velkymx @nolanpro should be better add a checkbox to enable the Password Grant type in the OAuth Client Screen, something like this:

image

similar to other OAuth applications:
image

velkymx:
I agree. @nolanpro please go ahead and make these changes.

caleeli:
@nolanpro, validation message is not shown to the end user when save client with no grants assigned.

image

nolanpro:
@caleeli ok added that

PR #2134, opened 2019-07-17
Alerts Not Timing Out

### LINKED ISSUE #2077: Alerts Not Timing Out
This is an extreme example, but shows how this element does not automatically close after X seconds displayed on the screen.

image

marcoAntonioNina:
resolves #2077 

- Change show time in alert of 60  to 5 seconds.
- add limit to only 6 items.


resources/js/app-layout.js
```diff
@@ -78,6 +78,10 @@ window.ProcessMaker.navbar = new Vue({
         alertDismissed(alert) {
             const index = this.alerts.indexOf(alert);
             index > -1 ? this.alerts.splice(index, 1) : null;
+            //remove old alerts
+            this.alerts = this.alerts.filter(item => {
+                return ((Date.now() - item.timestamp)/1000) < 5;
```
nolanpro: @marcoAntonioNina is this necessary now that we have reduced the default time and only allow 5 messages?

If you think it is still is needed, can we make this alertValue instead of hard-coded '5', and if alertValue is 0, then do not automatically remove it? This section of code is overwriting the showValue that can be passed in

marcoAntonioNina: @nolanpro you are right. I will modify to use the configured time.

PR #2136, opened 2019-07-17
Change process to designer, update breadcrumbs

velkymx:
The Process Tab does not only have processes.  It also has scripts, screens, vocabularies, and more.  All of these elements will eventually interact with Collections. Screens already do. So, we suggest renaming the process tab to "Designer".

I've updated the navbar and the breadcrumbs throughout the processes area.


ryancooley:
@velkymx Should we update routes as well or is that a future project/not necessary?

PR #2167, opened 2019-07-24
Add parent info to call activity

### LINKED ISSUE #2165: Include parent info in child process
In core: needed for ABE CallActivity

nolanpro:
Fixes #2165 

Needed for actions by email


ProcessMaker/Models/CallActivity.php
```diff
@@ -40,7 +40,20 @@ function ($self, TokenInterface $token, FlowInterface $sequenceFlow) {
...
-                $dataStore->setData($token->getInstance()->getDataStore()->getData());
+                $data = $token->getInstance()->getDataStore()->getData();
+
+                // Add info about parent
+                $data['_parent'] = [
+                    'process_id' => $token->getInstance()->process_id,
+                    'node_id' => $token->node_id,
+                ];
+
+                $configString = $this->getProperty('config');
```
velkymx: What happens if this isn't JSON? Is there always a config?

nolanpro: That should return null, thats why I do a check on the next line

nolanpro: Oh and config is the bpmn attribute 'pm:config' which is always json if it exists

PR #2181, opened 2019-07-26
Add new packages to config

velkymx:
Bumped modeler and screen builder


PMQABen:
Process Modeler works as expected again. Screen Builder is still full of bugs related to the Data Type field being added, but still seems to operate as expected.

PR #2292, opened 2019-08-15
Requests - Forms Detail Tab - Use historical data

### LINKED ISSUE #2291: Requests - Forms Detail Tab - Use historical data
The current Form Detail lists all the screens used in a request. The data displayed in those screens corresponds to the latest data of the request. 

Instead of showing the current request data, the screens should show  the data which was stored  at the moment the user submitted the form, so we need a mechanism to save the historical data as it was used and stored.

marcoAntonioNina:
Resolves #2291 

- add field data in Process Request Token for save the current data of screen.

Current data
image
Historical data in screens
image


marcoAntonioNina:
In the summary tab you will find the latest data. Only in the forms tab is the historical data

PR #2304, opened 2019-08-19
Changed string "Request All" to "All Requests"

PMQABen:


PMQABen:
I can throw the json updates into a separate commit for the translations package, but the middleware.php change would have to come through here.

PR #2308, opened 2019-08-21
Allow for mustache syntax before Evaluate Expression

### LINKED ISSUE #2307: Allow for mustache syntax before Evaluate Expression
image

marcoAntonioNina:
Resolves #2307 

- Add mustache syntax in evaluate expression

### DIFF
--- a/ProcessMaker/Models/FormalExpression.php
+++ b/ProcessMaker/Models/FormalExpression.php
@@ -2,6 +2,7 @@
 
 namespace ProcessMaker\Models;
 
+use Mustache_Engine;
 use ProcessMaker\Exception\ExpressionFailedException;
 use ProcessMaker\Exception\ScriptLanguageNotSupported;
 use ProcessMaker\Exception\SyntaxErrorException;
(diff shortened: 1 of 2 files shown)


ProcessMaker/Models/FormalExpression.php
```diff
@@ -43,6 +44,20 @@ private function initFormalExpression()
...
 
+    /**
+     * Parse mustache syntax
+     *
+     * @param string $expression
+     * @param array $data
+     *
+     * @return mixed
+     */
+    private function mustacheTimerExpression($expression, $data)
```
nolanpro: @marcoAntonioNina why name this with 'Timer'? just curious

ProcessMaker/Models/FormalExpression.php
```diff
@@ -64,7 +80,7 @@ private function evaluate(array $data)
         $encoder = isset(self::languages[$language][1]) ? self::languages[$language][1] : null;
         try {
             $values = $encoder ? $this->$encoder($data) : $data;
-            return $this->$evaluator->evaluate($this->getBody(), $values);
+            return $this->$evaluator->evaluate($body, $values);
```
nolanpro: @marcoAntonioNina can we add a test for this? I know there isn't a test file yet but it should be simple to create one for FormalExpression in tests/Model

ProcessMaker/Models/FormalExpression.php
marcoAntonioNina: fixed

PR #2334, opened 2019-08-28
add accept header to request

### LINKED ISSUE #2322: Unauthorized API
If an API endpoint is unauthorized, current behavior redirects to a login page. Instead it must return the status 401 (Unauthorized) with following JSON payload:

```
{
    "error": "Not Authorized"
}
```

sanjacornelius:
This PR adds the 'accept: application/json' header to the request. 

closes #2322

### DIFF
--- a/ProcessMaker/Exception/Handler.php
+++ b/ProcessMaker/Exception/Handler.php
@@ -83,6 +83,7 @@ class Handler extends ExceptionHandler
      */
     protected function unauthenticated($request, AuthenticationException $exception)
     {
+        $request->headers->set('Accept', 'application/json');
         if ($request->expectsJson()) {
             return response()->json(['error' => 'Unauthenticated.'], 401);
         }


sanjacornelius:
@velkymx did some more digging into this issue because the above 'fix'  broke another test as well as I learned it's bad practice.

Within the original test that was created
```
public function testAuthGetAllSources()
    {
        $response = $this->get('/api/1.0'.self::API_URL);
        $response->assertUnauthorized();        
    }
```
`$this->get()`  is being used which is a web call and doesn't set the right headers, which why the test is failing and you are being redirected to the login page.

Instead, use the`$this->json()` method. This will set the correct headers for the API call and return 401.

```
public function testAuthGetAllSources()
    {
        $response = $this->json('GET', self::API_URL);
        $response->assertUnauthorized();        
    }
```

PR #2347, opened 2019-08-30
Use button type="button"

### LINKED ISSUE #2311: Form Submit
Where ever we use axios we need to remove the default form submit function. (enter to submit, etc)

caleeli:
Resolves #2311 

This PR changes button type="submit" for screens that use axios to send the data.


velkymx:
@PMQABen can you confirm this works?

PMQABen:
❌ Still occurs when renaming an existing category for:
* Processes
* Scripts
* Screens

caleeli:
Fixed. I found the problem in the blades that use the laravel form helpers {{ Form::open() }}

PMQABen:
✅

PR #2355, opened 2019-09-02
Change breadcrumbs to use display flex 

### LINKED ISSUE #2348: Flex Box Layout
Update the main layout to correctly use flexbox.

https://getbootstrap.com/docs/4.0/utilities/flex/

Current issue being addressed is the scrolling is overlapping the scroll bar.

caleeli:
Resolves #2348 

This PR includes:

- Move the breadcrumbs and use display flex.
- Remove old screen


resources/views/admin/auth-clients/index.blade.php
```diff
@@ -8,11 +8,13 @@
     @include('layouts.sidebar', ['sidebar'=> Menu::get('sidebar_admin')])
 @endsection
 
-@section('content')
+@section('breadcrumbs')
```
velkymx: breadcrumbs section should be defined in the layout.blade not in the individual blades

caleeli: @yield('breadcrumbs') is already defined in the layout.blade and was used in modeler, screen builder, script editor, ...

PR #2437, opened 2019-10-08
Improve download speed

JCook21:
These changes normalize versions of some dependencies between modeler and process maker. This allows the modeler package to not have to include these dependencies when bundling the package for production, which results in a roughly 50% reduction in the package size (in some cases libraries like jointjs were being included in the modeler and process maker packages).

I also moved a number of third party modules out into a vendor.js file using `mix.extract`. This should mean that file can be better cached by browsers.

These changes work independently but work well with upcoming changes to the modeler to reduce bundle size there.


package.json
```diff
@@ -35,7 +35,6 @@
     "jest": "^22.4.4",
     "jquery": "^3.1.1",
     "laravel-mix": "2.1.*",
-    "lodash": "^4.17.10",
```
chipit24: Since you moved `lodash` over to `dependencies`, I think you may as well move `jquery`, `moment`, `moment-timezone` and `vue` over as well. Though for this repo it doesn't really matter where it goes.

JCook21: Hmm, I think I moved lodash over as we have it as a main dep modeler while it was in dev here. I can look at moving the others over later if need be...

JCook21: I want to try to update the deps in this repo so maybe I can come back to this then.

PR #2454, opened 2019-10-15
Implement multiple categories for Process/Screens/Scripts

### LINKED ISSUE #2428: Allow items (screens, scripts, processes) to belong to multiple categories
- [x] Update model relations and add migration
- [x] Migrate old category ids and remove _category_id columns
- [ ] Update API endpoints
- [ ] Update UI to multiselect

caleeli:
Resolves #2428 

This PR includes:

- Changes in Migrations, Model, Views, Vue components to support multiple categories for: Processes, Screens and Scripts

### DIFF
--- a/ProcessMaker/Models/CategoryAssignment.php
+++ b/ProcessMaker/Models/CategoryAssignment.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace ProcessMaker\Models;
+
+use Illuminate\Database\Eloquent\Relations\Pivot;
+
+class CategoryAssignment extends Pivot
+{
+    public function setMorphType($type)
+    {
+        return $this;
+    }
+
+    public function setMorphClass($class)
+    {
+        return $this;
+    }
+
+    public static function boot()
+    {
+        parent::boot();
+        static::saved(function ($pivot) {
+            switch ($pivot->assignable_type) {
+                case Process::class:
+                    //$pivot->assignable->process_category_id = $pivot->category->getKey();
+                    break;
+                case Script::class:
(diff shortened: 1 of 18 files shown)


nolanpro:
@caleeli i'm getting this when running `php artisan migrate` after running migrate:fresh on develop then switching to this branch
```
Symfony\Component\Debug\Exception\FatalThrowableError  : Argument 1 passed to Illuminate\Database\Eloquent\Relations\BelongsToMany::save() must be an instance of Illuminate\Database\Eloquent\Model, array given, called in /home/vagrant/processmaker/database/migrations/2019_09_30_192339_create_category_assignments_table.php on line 33
```

nolanpro:
I like the use of a Pivot model here, didn't know you could do that!

- There isn't a migration to remove existing `*_category_id` from the process/script/screen tables and the `->category()` method still exists in those models. Is this intentional?
- So we're not making categories fully polymorphic? We're keeping the `*_categories` tables for now?
- Would be good to have some new tests or modify existing category tests for this new schema.

We *might* be able to keep the category() method (to ease the transition in packages that expect that method to exist) by doing something crazy like this:
```
public function category()
    {
        $query = $this->categories()->orderBy('is_system', 'desc')->limit(1);
        return new \Illuminate\Database\Eloquent\Relations\BelongsTo(
            $query, $this, .....
        )
    }
```
I have no idea if that would work. Just a thought. But if it did work we could get rid of the *_category_id columns

caleeli:
@nolanpro, thanks for the report, the migration issue was fixed

caleeli:
I intentionally keep the `*_category_id` fields to keep the sort functionality by category.

PR #2481, opened 2019-10-24
Update system resource/asset management

### LINKED ISSUE #2451: System Asset Control
If an asset is part of the "system" category, then users should not be allowed to edit the asset. We should be encrypting the content - maybe using the new encryption train to prevent tampering.

nolanpro:
Fixes #2451 
- Locks down viewing/editing resources (Process, ProcessRequest, ProcessRequestToken, Scripts, Screens) when they belong to a system category
- Encrypts specific fields in those models when saved
- Utilizes cache since determining if an asset belongs to a system category can be slow


velkymx:
@nolanpro this is failing unit tests.

nolanpro:
@velkymx this passes phpunit now. The other is a devops problem

PR #2485, opened 2019-10-25
Add default value to specialAssignmentsListGetter

### LINKED ISSUE #2483: Assignment Rules not loading for task in modeler
The _Assignment Rules_ accordion is empty for Task, and selecting the Task produces the following console error:

```
Error in render: "SyntaxError: Unexpected token u in JSON at position 0"
```

The exception is thrown from the following block of code:
https://github.com/ProcessMaker/processmaker/blob/cb59bfba875ff7f1d57e8c141b9c8d12574f7ed7/resources/js/processes/modeler/components/inspector/TaskAssignment.vue#L188-L191

tuantruong625:
This PR fixes the console error `Error in render: "SyntaxError: Unexpected token u in JSON at position 0"` 
The Assignment Rules accordion renders but when creating a new rule a console error is thrown.

```
TypeError: Cannot read property 'content' of undefined
    at VueComponent.saveSpecialAssignment (initialLoad.js:336498)
    at invokeWithErrorHandling (vendor.js:97653)
    at HTMLButtonElement.invoker (vendor.js:97978)
    at HTMLButtonElement.original._wrapper (vendor.js:103331)
```

Fixes #2483

### DIFF
--- a/resources/js/processes/modeler/components/inspector/TaskAssignment.vue
+++ b/resources/js/processes/modeler/components/inspector/TaskAssignment.vue
@@ -186,8 +186,8 @@
         return this.typeAssignmentExpression === "group";
       },
       specialAssignmentsListGetter () {
-        const value = _.get(this.node, "assignmentRules", '[]');
-        return JSON.parse(value) || [];
+        const value = this.node.get('assignmentRules') || '[]';
+        return JSON.parse(value);
       },
     },
     methods: {
--- a/resources/js/processes/modeler/components/inspector/TaskAssignment.vue
+++ b/resources/js/processes/modeler/components/inspector/TaskAssignment.vue
@@ -186,7 +186,7 @@
         return this.typeAssignmentExpression === "group";
       },
       specialAssignmentsListGetter () {
-        const value = _.get(this.node, "assignmentRules");
+        const value = _.get(this.node, "assignmentRules", '[]');
         return JSON.parse(value) || [];
       },
     },


lisa-fehr:
Looking good so far

resources/js/processes/modeler/components/inspector/TaskAssignment.vue
```diff
@@ -186,7 +186,7 @@
         return this.typeAssignmentExpression === "group";
       },
       specialAssignmentsListGetter () {
-        const value = _.get(this.node, "assignmentRules");
+        const value = _.get(this.node, "assignmentRules", '[]');
         return JSON.parse(value) || [];
```
JCook21: Is the or statement needed on this line now? Since we have a default value on line 189 can we remove the duplication on this line?

tuantruong625: Removing the `|| []` throws the same error again.

chipit24: I think the `_.get` on line  189 is redundant. What if you change it to:
```js
const value = this.node.get('assignmentRules') || '[]';
```

Do you then still need the `|| []`?

According to [the docs](https://lodash.com/docs/4.17.15#get),`_.get` will only fall back to the default value if it resolves to `undefined`, but the code I suggested above will fall back to the `'[]'` value if it's an empty string or `null` as well.

tuantruong625: `const value = this.node.get('assignmentRules') || '[]';` still produces the same error.

PR #2520, opened 2019-11-13
Fix new scripts creating weird assets

### LINKED ISSUE #2515: New Scripts Creating Weird Assets
Ensure your VM is running PM4 core, no extra packages, and that the database has been reset ({{art migrate:fresh --seed --force}})
1. Check /admin/users to ensure there is only one user, "admin admin"
1. Go to Designer -> Scripts
1. Create a category, then create a script (any language)
1. Go back to /admin/users and refresh the page

**Result:**
A new user has been created
https://www.dropbox.com/s/4m9ad1tutj3819t/new-script-creating-user.mp4?dl=0

**Note:**
Looks like it's also creating new Process categories

nolanpro:
Fixes #2515 

Get process request attributes without using a factory

### DIFF
--- a/ProcessMaker/Http/Controllers/Process/ScriptController.php
+++ b/ProcessMaker/Http/Controllers/Process/ScriptController.php
@@ -2,7 +2,8 @@
 
 namespace ProcessMaker\Http\Controllers\Process;
 
-use Illuminate\Support\Facades\Auth;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Schema;
 use ProcessMaker\Http\Controllers\Controller;
 use ProcessMaker\Models\ProcessRequest;
 use ProcessMaker\Models\Script;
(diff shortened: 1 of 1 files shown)


ProcessMaker/Http/Controllers/Process/ScriptController.php
```diff
@@ -48,18 +49,33 @@ public function edit(Script $script, User $users)
...
-                'process_id' => 1,
-                'process_collaboration_id' => 1,
-            ]),
+            '_request' => $processRequestAttributes
         ];
-
         return view('processes.scripts.builder', compact('script', 'scriptFormat', 'testData'));
     }
+
+    private function getProcessRequestAttributes()
```
velkymx: Does this need to be broken out into its own method? I feel like the problem lay in the factory not the Schema. If we call that factory for something else will the problem come back?

PR #2573, opened 2019-11-15
Fix New Script modal not resetting

sanjacornelius:
Converted standard modal to vue bootstrap modal.


resources/views/processes/scripts/list.blade.php
```diff
@@ -26,159 +26,86 @@ class="fas fa-plus"></i>
...
+                    @endif
+                </template>
+            </b-modal>
+        @endcan
+    </div>
 
 @section('js')
     <script src="{{mix('js/processes/scripts/index.js')}}"></script>
-
-    @can('create-scripts')
```
velkymx: Please put this logic back. The point, while not as compact, is to make it easier to work with by seeing the Vue object via the blade and not having to hunt for it.

PR #2582, opened 2019-11-20
Add default category

### LINKED ISSUE #2552: Add default category
When the system is initially installed all categories need to include a default category "Uncategorized". Add to the seeder

marcoAntonioNina:
Resolves #2552 

- Add seeder for:
- processes
- screens
- scripts

### DIFF
--- a/database/seeds/CategorySystemSeeder.php
+++ b/database/seeds/CategorySystemSeeder.php
@@ -14,8 +14,6 @@ class CategorySystemSeeder extends Seeder
      */
     public function run()
     {
-        $name = __('Uncategorized');
-
         factory(ScreenCategory::class)->create([
             'name' => __('Uncategorized'),
             'status' => 'ACTIVE',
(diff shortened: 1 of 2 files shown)


database/seeds/CategorySystemSeeder.php
```diff
@@ -0,0 +1,35 @@
...
+class CategorySystemSeeder extends Seeder
+{
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
+        $name = __('Uncategorized');
```
velkymx: This shouldn't translate...

database/seeds/DatabaseSeeder.php
```diff
@@ -21,7 +21,8 @@ public function run()
             ProcessSystemCategorySeeder::class,
             GroupSeeder::class,
             ScreenTypeSeeder::class,
-            ScreenSystemSeeder::class
+            ScreenSystemSeeder::class,
```
velkymx: Do we need all of these seeders with this change?

marcoAntonioNina: The default screen is used in Feature interstitial

marcoAntonioNina:
@velkymx comments fixed.

PR #2591, opened 2019-11-22
Fix versioning of screens and scripts in process execution

### LINKED ISSUE #2457: Problems with the displayed screen in "Forms" tab
There are two problems with the displayed data in the historical  "Forms" tab.

1) It doesn't show the screen that was used when the task was submitted. It shows the latest screen. 
2) If the screen used by the task is set to other one, this screen is used in the "Forms" tab.

Both behaviors can be seen in the attached videos.

Peek 2019-10-15 16-34
Peek 2019-10-15 16-08

caleeli:
Resolves #2457 

This PR includes:

- Scripts, Screen and Processes has always one version
- Move versioning creation to a trait "HasVersioning"
- Fix screens history to show the version in which the form was saved.
- Add version column to the tokens


caleeli:
> I'm not able to view the form on the form tab. When I click on the details icon, i get "[Vue warn]: Error in render: "TypeError: Cannot read property 'filter' of undefined"
> When I click on the print icon, it opens a new tab to `http://pm.local.processmaker.com/requests/1/screen/3` that results in a 404

Fixed issue saving the versions (json fields were being saved as json strings) that cause the problem when loading the form config.

PR #2616, opened 2019-11-29
Export 500

### LINKED ISSUE #2540: Export 500
When a name of an element contains a forward or back slash, the export fails.  We should only support alpha-numeric values in names.

danloa:
Resolves #2540 

- As Laravel's default validator are limited, a new validator that accepts just letters (UTF-8) + spaces + dashes (_ -)  was added. 
- The validations were added to all models that have names or titles.


velkymx:
@marcoAntonioNina make sure you're only committing files you actually changed. There are some formatting issues here that shouldn't have been committed.

PR #2618, opened 2019-11-29
Export with All Caps

### LINKED ISSUE #2543: Export with All Caps
If you export a process with all caps (NAMEHERE) it will be exported as n_a_m_e_h_e_r_e should just use the upper case provided.

danloa:
Resolves #2543 

The problem was caused for the use of the deprecated function snake_case. It was removed and now the exported file respects the original casing and formatting the process and screen names

### DIFF
--- a/ProcessMaker/Http/Controllers/Process/ScreenController.php
+++ b/ProcessMaker/Http/Controllers/Process/ScreenController.php
@@ -95,7 +95,7 @@ class ScreenController extends Controller
      */
     public function download(Screen $screen, $key)
     {
-        $fileName = snake_case($screen->title) . '.json';
+        $fileName = trim($screen->title) . '.json';
         $fileContents = Cache::get($key);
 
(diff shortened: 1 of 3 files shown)


ProcessMaker/Http/Controllers/Process/ScreenController.php
```diff
@@ -95,7 +95,7 @@ public function import(Screen $screen)
      */
     public function download(Screen $screen, $key)
     {
-        $fileName = snake_case($screen->title) . '.json';
+        $fileName = trim($screen->title) . '.json';
```
velkymx: I think the SLUG method would be a better choice.

https://laravel.com/docs/5.7/helpers#method-str-slug

danloa: The problem with the slug method is that it lowercases the name. So MyProcess will be exported as "myprocess", is this ok?

PR #2628, opened 2019-12-03
Print screen with multiple pages

### LINKED ISSUE #2561: Print Multi-Page Forms
In the Request summary there is the option to print multi-page forms, but you can only see page one. Instead show the other pages stacked (in order) below the first page.

caleeli:
Resolves #2561 

This PR includes the following changes:

- For a screen with multiple pages, includes a preview of each of the pages
- Print all the pages of a screen
- Only print the pages that are navigable (if the page can not be reached ex. a used to fill a grid, they will not be printed)

image

image


caleeli:
Fixed issue related to wrong ID when printing screens, and getting the list of printable pages.

PR #2658, opened 2019-12-11
Added Open links to Task Screens and Scripts

lisa-fehr:
Closes https://github.com/ProcessMaker/modeler/issues/942

### DIFF
--- a/resources/js/processes/modeler/components/inspector/ScreenSelect.vue
+++ b/resources/js/processes/modeler/components/inspector/ScreenSelect.vue
@@ -25,10 +25,10 @@
         <small v-if="helper" class="form-text text-muted">{{ $t(helper) }}</small>
         <a
                 v-if="content.id"
-                :href="`/designer/screens/${content.id}/edit`"
+                :href="`/designer/screen-builder/${content.id}/edit`"
                 target="_blank"
         >
-            Open Screen
+            {{ $t('Open Screen') }}
             <i class="ml-1 fas fa-external-link-alt"/>
         </a>
     </div>
--- a/resources/js/processes/modeler/components/inspector/ScreenSelect.vue
+++ b/resources/js/processes/modeler/components/inspector/ScreenSelect.vue
@@ -23,6 +23,14 @@
         </multiselect>
         <small v-if="error" class="text-danger">{{ error }}</small>
         <small v-if="helper" class="form-text text-muted">{{ $t(helper) }}</small>
+        <a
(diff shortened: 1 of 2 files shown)


resources/js/processes/modeler/components/inspector/ScreenSelect.vue
```diff
@@ -23,6 +23,14 @@
         </multiselect>
         <small v-if="error" class="text-danger">{{ error }}</small>
         <small v-if="helper" class="form-text text-muted">{{ $t(helper) }}</small>
+        <a
+                v-if="content.id"
+                :href="`/designer/screens/${content.id}/edit`"
+                target="_blank"
+        >
+            Open Screen
```
chipit24: Are translations available for this link text? Could it be wrapped in `$t`?

resources/js/processes/modeler/components/inspector/ScreenSelect.vue
```diff
@@ -23,6 +23,14 @@
         </multiselect>
         <small v-if="error" class="text-danger">{{ error }}</small>
         <small v-if="helper" class="form-text text-muted">{{ $t(helper) }}</small>
+        <a
+                v-if="content.id"
+                :href="`/designer/screens/${content.id}/edit`"
```
chipit24: Should this go to `/designer/screen-builder/${content.id}/edit` instead? The current URL seems to open a page to edit only the name, description, and category, and not the screen builder.

resources/js/processes/modeler/components/inspector/ScreenSelect.vue
lisa-fehr: No translations. Should I wrap it?

resources/js/processes/modeler/components/inspector/ScreenSelect.vue
lisa-fehr: That is the link I was taken to. Let me check.

resources/js/processes/modeler/components/inspector/ScreenSelect.vue
lisa-fehr: "Open Process" isn't wrapped. So I followed that.

chipit24: I should have wrapped "Open Process". I think it makes sense to, as everything seems to be translated. Worst case is the translation doesn't exist and it shows the text as-is.

Issue #2702, opened 2020-01-03
Add missing translations


lisa-fehr:
Go through modeler and add missing translations.

<img width="479" alt="Screen Shot 2020-01-03 at 1 53 28 PM" src="https://user-images.githubusercontent.com/6653340/71742674-8c936800-2e30-11ea-9403-78335fd02ca4.png">

PR #2703, opened 2020-01-03
Add missing translations

lisa-fehr:
Closes: #2702 

Altered translation with colon. Wasn't parsing correctly, might be possible to escape the colon.
Adding missing translations.

`CLOSE` may still show up as a missing translation when adding a Script Task and expanding the Script Configurator Editor. This will eventually go away once this pr has a release # https://github.com/ProcessMaker/screen-builder/pull/491


resources/lang/en.json
```diff
@@ -127,6 +128,7 @@
   "Client ID": "Client ID",
   "Client Secret": "Client Secret",
   "Close": "Close",
+  "CLOSE": "CLOSE",
```
chipit24: I don't think we should have different translation for an all-caps version of the word "close". I think the translation above this one ("Close") should be used, and a style of `text-transform: capitalize;` should be applied to it. There is a bootstrap helper for this too: https://getbootstrap.com/docs/4.0/utilities/text/#text-transform.

```
<p class="text-lowercase">Lowercased text.</p>
<p class="text-uppercase">Uppercased text.</p>
<p class="text-capitalize">CapiTaliZed text.</p>
```

resources/lang/en.json
```diff
@@ -363,13 +366,16 @@
   "Group Permissions Updated Successfully": "Group Permissions Updated Successfully",
   "Group Permissions": "Group Permissions",
   "Group": "Group",
+  "group": "group",
```
chipit24: Same thing here regarding capitalization. I think capitalization should be handled with CSS and whether certain words are capitalized or not can be controlled with the bootstrap text helper classes.

resources/lang/en.json
lisa-fehr: I thought I found `CLOSE` in screen builder, but when I change it and link it into processmaker it doesn't change.

resources/lang/en.json
lisa-fehr: When searching, I have trouble finding where `Group` and `group` occur. Some are headers and some are select options. Switching all instances of `Group` might be tricky for me. I'm not sure when/where the `group` translation was introduced.

velkymx: This is fine for now. We'll circle back if becomes an issue.

PR #2706, opened 2020-01-06
Fix Confirm Modal test

lisa-fehr:
Span was switched to div.
Screen Shot 2020-01-06 at 9 46 12 AM

### DIFF
--- a/tests/js/components/ConfirmModal.test.js
+++ b/tests/js/components/ConfirmModal.test.js
@@ -41,7 +41,7 @@ describe('Component Confirm Modal', () => {
       }
     });
     expect(wrapper.find('.modal-title').text()).toBe(title);
-    expect(wrapper.find('.modal-body span').text()).toBe(message);
+    expect(wrapper.find('.modal-body div').text()).toBe(message);
 
   });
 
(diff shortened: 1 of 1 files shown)


tests/js/components/ConfirmModal.test.js
```diff
@@ -41,7 +41,7 @@ describe('Component Confirm Modal', () => {
       }
     });
     expect(wrapper.find('.modal-title').text()).toBe(title);
-    expect(wrapper.find('.modal-body span').text()).toBe(message);
+    expect(wrapper.find('.modal-body div').text()).toBe(message);
```
chipit24: As it stands, `.modal-body div` is not a descriptive selector and seems quite brittle as this test would fail if the nesting or tag name changes again. Do you think it would be worth adding a data-test attribute to the message and using that in this test? Also, I wonder if `.text()` can be called directly on `.modal-body` to get the text content without having to reach into whatever child nodes it has.

lisa-fehr: @chipit24 `.modal-body` alone also works.

PR #2722, opened 2020-01-09
Merge breadcrumb nav and toolbar

tuantruong625:
The purpose of this PR is to merge the breadcrumb navigation with the toolbar into a single bar.

**Note**
- Replace breadcrumbs with modeler toolbar which contains, breadcrumbs and toolbar

Associated PR in Modeler
https://github.com/ProcessMaker/modeler/pull/1010

### DIFF
--- a/resources/views/processes/modeler/index.blade.php
+++ b/resources/views/processes/modeler/index.blade.php
@@ -57,10 +57,10 @@ div.main {
   }
   const warnings = @json($process->warnings);
  
-  window.ProcessMaker.EventBus.$on('modeler-start', ({ loadXML, addWarnings, breadcrumbs }) => {
+  window.ProcessMaker.EventBus.$on('modeler-start', ({ loadXML, addWarnings, addBreadcrumbs }) => {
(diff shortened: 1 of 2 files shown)


resources/views/processes/modeler/index.blade.php
```diff
@@ -39,30 +26,41 @@
...
+    xml: @json($process->bpmn),
+    processName: @json($process->name),
   }
   const warnings = @json($process->warnings);
-  window.ProcessMaker.EventBus.$on('modeler-start', ({ loadXML, addWarnings }) => {
+ 
+  window.ProcessMaker.EventBus.$on('modeler-start', ({ loadXML, addWarnings, breadcrumbs }) => {
     loadXML(window.ProcessMaker.modeler.xml);
     addWarnings(warnings || []);
+    breadcrumbs(breadcrumbData || []);
```
chipit24: Do you think a name like `addBreadcrumbs` would be more readable? I think it will keep it consistent with the naming for the function above it (`addWarnigns`) and also better communicate what the function is actually doing.

PR #2728, opened 2020-01-10
Updated dependencies

velkymx:
Update for package json and composer


package.json
```diff
@@ -16,7 +16,7 @@
   "devDependencies": {
     "@vue/test-utils": "^1.0.0-beta.30",
     "accounting": "^0.4.1",
-    "axios": ">=0.19.0",
```
chipit24: Why was the version specifier change to `>=`? This would include future breaking changes as well. The `^` will ensure a compatible version is installed, and I think it can be left the way it is.

velkymx: I didn't do it specifically. I did npm update and that was the change.

chipit24: Interesting, I didn't think it would change it so drastically. In either case, the lockfile will keep the version consistent so it's not a problem.

PR #2765, opened 2020-01-17
Restored Allowed Users, Allowed Groups and Whitelist fields on Intermediate Catch Event

lisa-fehr:
Closes https://github.com/ProcessMaker/modeler/issues/1032


lisa-fehr:
Investigating error (don't see the duplicate):
Screen Shot 2020-01-17 at 3 53 50 PM

PR #2794, opened 2020-01-28
Add more descriptive expresion syntax error and redirect user on request error

### LINKED ISSUE #2788: Failed Request Messaging
Steps to Reproduce:

1. Create a process with a exclusive gateway
2. set wrong expression in the outgoing sequences flows
3. Run a case through the gateway
4. See a 500 Error (Request failed with status code 500)

Edit: Alan mentioned this happens when the 2 expressions are the same

**CURRENT BEHAVIOR:**

"Request failed with status code 500" error is shown when there are wrong expressions set for an exclusive gateway. Also note that once the issue happened you can't continue with the request even after the issue got fixed from the designer.

**Expected Behavior:**

500 error is too generic, users should see something more specific like "Wrong expressions set for gateway" or "Any expression was evaluated as true, contact your administrator" or something else. You should be able to continue with the request once the issue got fixed from the designer

nolanpro:
Fixes #2788 

- Add more descriptive message for expression syntax errors
- Redirect the user to the request page to view said error instead of just flashing it quickly


nolanpro:
Waiting for translations

PR #2796, opened 2020-01-28
Filter screens listed in nested screens control

### LINKED ISSUE #2793: Nested Screen implementation
Ability to render/nest a screen (display can only nest other display forms, a input screen can nest a display/input screen).

caleeli:
Resolves #2793 

This PR includes:

- Add the PMQL parameter to screen index API

Required by: Nested screen feature

### DIFF
--- a/ProcessMaker/Models/Screen.php
+++ b/ProcessMaker/Models/Screen.php
@@ -155,7 +155,7 @@ class Screen extends Model
     }
 
     /**
-     * PMQL field alias (created = created_at)
+     * PMQL field (id = screens.id)
      *
      * @return string
      */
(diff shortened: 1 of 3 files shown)


ProcessMaker/Models/Screen.php
```diff
@@ -151,4 +153,14 @@ public function renderComponent()
         }
         return 'task-screen';
     }
+
+    /**
+     * PMQL field alias (created = created_at)
```
danloa: It seems that this comment should have
"PMQL field alias (id = screens.id)"

caleeli: Fixed

PR #2803, opened 2020-01-30
Package: Dynamic Home Page

### LINKED ISSUE #2758: Package: Dynamic Home Page
Allow user to specify default login page. Some sort of “Make my home” button or icon for major areas so that when a user logs in again, it redirects immediately to that page.

We need this to be internally reportable (via a SQL query) so we can determine the use of this feature so we can expand later

https://github.com/ProcessMaker/package-dynamic-homepage

danloa:
Resolves #2758 

This PR adds "hooks" so that any package can add functionalities to the edit profile and login process flow. The way to add is the same to the one used in the vocabularies package.


ProcessMaker/Http/Controllers/Auth/LoginController.php
```diff
@@ -39,6 +41,10 @@ public function __construct()
         $this->middleware('guest')->except(['logout', 'keepAlive']);
     }
 
+    public function redirectTo() {
+        return "xxx";
```
velkymx: What does this do?

ProcessMaker/Http/Controllers/ProfileController.php
```diff
@@ -6,10 +6,13 @@
 use ProcessMaker\Models\User;
 use ProcessMaker\Models\JsonData;
 use ProcessMaker\i18nHelper;
+use ProcessMaker\Traits\HasControllerAddons;
```
velkymx: Does it make sense to add this here? Maybe in the parent controller would be a better fit?

resources/views/profile/edit.blade.php
```diff
@@ -76,8 +76,18 @@
 	<script src="{{mix('js/admin/profile/edit.js')}}"></script>
 
     <script>
+        var addons = [];
```
velkymx: I feel like this would be better in the layout blade.

ProcessMaker/Http/Controllers/Auth/LoginController.php
danloa: By error testing code was left... Fixed

resources/views/profile/edit.blade.php
danloa: In principle I was trying to limit the scope of addons. I moved this code to the layout and base controller so that any page can have addons now.

ProcessMaker/Http/Controllers/ProfileController.php
danloa: In principle I was trying to limit the scope of addons. I moved this code to the layout and base controller so that any page can have addons now.

PR #2813, opened 2020-02-03
Check if container items are arrays and convert if not

### LINKED ISSUE #2812: Add support for non-array container items in screen builder
Required for https://github.com/ProcessMaker/screen-builder/pull/527

Container items are expected to be an array of items (each column contains an array of items). This is no longer the case with Loop containers.

nolanpro:
Fixes #2812 

Required for ProcessMaker/screen-builder#527

Container items are expected to be nested arrays of items (for Multi Column containers, each column contains an array of items). This is no longer the case with Loop containers. A Loop container is a single array of items. There are a number of validation checks here in core that rely on this nested array structure (I don't know why these checks are not in screen-builder)

The simple solution is to check and see if 'items' is an array, and if it is not, wrap it in a single-item array.

### DIFF
--- a/resources/js/processes/screen-builder/screen.vue
+++ b/resources/js/processes/screen-builder/screen.vue
@@ -306,6 +306,10 @@ import formTypes from "./formTypes";
       },
       getValidationErrorsForItems(items, page) {
         const validationErrors = [];
+            
+        if (!Array.isArray(items)) {
+          items = [items];
+        }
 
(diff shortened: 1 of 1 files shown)


velkymx:
@nolanpro What is this PR fixing? Can you add more details?

nolanpro:
@velkymx updated the PR description

PR #2919, opened 2020-03-03
Fix Environment Variable Text Limit

### LINKED ISSUE #2875: Environment Vars Text Limit
The environment variable shouldn't have a text limit. I need to be able to store longer data, such as bearer tokens, and currently I am limited.

sanjacornelius:
<h2>Changes</h2>

- Change the env variable value Input to a Textarea, allowing for displaying longer data. 

- Convert the env variable value column type from 'TEXT' to 'LONGTEXT' in the database allowing storing of data larger than 64kb. 

https://www.loom.com/share/2697f4c87e9b4545a52e333ff90d46d7

<h2>To Test</h2>

- Create an Env Variable with a value longer than 64kb and save

**Expected Result**
Env variable is successfully created.

closes #2875

### DIFF
--- a/database/migrations/2020_03_03_181316_update_env_var_value_column.php
+++ b/database/migrations/2020_03_03_181316_update_env_var_value_column.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class UpdateEnvVarValueColumn extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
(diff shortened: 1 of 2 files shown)


johnblackspear:
Works well from my testing, and the migration works (I tested rollback too, that works well).


UI changes are good, too:




I've been able to store and retrieve an 80kb value (the previous column, `text` maxed out at 64kb) so the longtext fix works.

I've one question - what gets stored in the `notifications` table on script execution? I haven't been able to get my notifications working so I'm not sure that my question even makes sense. But if (maybe) the script's full output goes into `notifications`'s `data` column that might need to be changed, too.

I'm asking because I saw the `sendResponse` code in `ExecuteScript`, please ignore if I'm talking nonsense.

PR #2955, opened 2020-03-12
Remove task config override

chipit24:
This change is required for the documentation feature to work correctly: https://github.com/ProcessMaker/modeler/pull/1137. The code for setting the default Task Assignment data has been moved to that PR.

### DIFF
--- a/resources/js/processes/modeler/initialLoad.js
+++ b/resources/js/processes/modeler/initialLoad.js
@@ -71,13 +71,6 @@ let nodeTypes = [
 ProcessMaker.nodeTypes.push(startEvent);
 ProcessMaker.nodeTypes.push(...nodeTypes);
 
-// Set default properties for task
-task.definition = function definition(moddle) {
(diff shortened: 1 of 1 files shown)


lisa-fehr:
Not sure on this one. Is there another way to select the first option? Or should the first option be empty?

Used to pre-select `Requester` (and it does still seem to check Requester without showing it):
Screen Shot 2020-03-12 at 2 03 29 PM
Screen Shot 2020-03-12 at 2 03 42 PM

chipit24:
@lisa-fehr Good point; this should not be merged until the branch in https://github.com/ProcessMaker/modeler/pull/1137 is merged. I'll add a note to this PR. If you test this without linking in the modeler PR then you get that behaviour; but when you `npm link` that branch it works as it did before.

velkymx:
Merge after https://github.com/ProcessMaker/processmaker/pull/2987

PR #3011, opened 2020-04-01
Hooks package version

marcoAntonioNina:
**Changes**

- Add Component Menu for Screen and Script Builder
- add hooks for packages
- add way for register js for Script builder


chipit24:
@marcoAntonioNina We can review the code but we're not completely sure what the hooks package is and how to property test this.

Can you please link us to an issue or provide a description of what this PR is suppose to change and how we can test it.

resources/js/processes/modeler/components/ModelerApp.vue
```diff
@@ -93,13 +100,24 @@
...
+      this.dataXmlSvg.svg = svg;
+
+      this.externalEmit.forEach(item => {
+        window.ProcessMaker.EventBus.$emit(item);
+      })
+      if (this.saveEmitDefault) {
+        window.ProcessMaker.EventBus.$emit('modeler-save');
+      }
+    },
+    saveBpmn(resolve, reject) {
```
johnblackspear, line 114: This isn't a promise - these are `onSuccess` and `onError` handlers that can optionally be passed in.

Could we change the naming from `resolve` to `onSuccess` and `reject` to `onError` since using the promise terminology misinforms?

resources/js/components/Menu.vue
```diff
@@ -0,0 +1,148 @@
...
+      this.changeOptions(this.section.right);
+    },
+    changeOptions(data) {
+      data.forEach(item => {
+        item = Object.assign(item, this.changeItems[item.id]);
+        if (item.type === 'group') {
+          this.changeOptions(item.items);
+        }
+      });
+    },
```
chipit24, line 142: These two methods are not called anywhere in this component. It seems they are set up to only be called externally, through a `ref`. This poses a problem, since these methods are part of this components internal state, and it will be assumed they won't be called from outside the component directly. This makes refactoring very difficult. But also makes understanding this code more difficult too, as now we have to jump between two components to fully understand how this is suppose to work.

From [the docs](https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements), regarding refs:
> It is only meant as an escape hatch for direct child manipulation

Please refactor this to use the component's _public_ API, that is: ↓**props down**↓, ↑**events up**↑. 

`section` is already a computed property, so you can delete these two methods, and wherever you need to change a group or button, change it in the options that are passed down through the `options` prop.

resources/js/processes/scripts/components/ScriptEditor.vue
```diff
@@ -175,9 +186,15 @@ export default {
...
   components: {
-    MonacoEditor
+    MonacoEditor,
+    MenuScript,
   },
   mounted() {
+    ProcessMaker.EventBus.$emit("script-builder-init", this);
+    ProcessMaker.EventBus.$on("save-script", (resolve, reject) => {
+      this.save(resolve, reject);
+    });
```
chipit24, line 196: Same here re: resolve and reject.

resources/views/processes/screens/edit.blade.php
```diff
@@ -58,6 +79,7 @@
     <script>
         new Vue({
             el: '#editGroup',
+            mixins: addons,
```
chipit24, line 82: Is this right? Mixins should be an array of JS objects; I don't see where `addons` is defined.

resources/views/processes/scripts/edit.blade.php
```diff
@@ -70,6 +100,7 @@
     <script>
       new Vue({
         el: '#editScript',
+        mixins: addons,
```
chipit24, line 103: Same here.

resources/js/processes/modeler/components/ModelerApp.vue
```diff
@@ -93,13 +100,24 @@
...
     },
-    saveBpmn({xml, svg}) {
+    actionSaveBpmn ({xml, svg}) {
+      this.dataXmlSvg.xml = xml;
+      this.dataXmlSvg.svg = svg;
+
+      this.externalEmit.forEach(item => {
+        window.ProcessMaker.EventBus.$emit(item);
+      })
+      if (this.saveEmitDefault) {
```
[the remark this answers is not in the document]
marcoAntonioNina: renamed to saveProcess

resources/js/processes/modeler/components/ModelerApp.vue
```diff
@@ -93,13 +100,24 @@
...
+    },
+    saveBpmn(resolve, reject) {
       const data = {
         name: this.process.name,
         description: this.process.description,
         task_notifications: this.getTaskNotifications(),
-        bpmn: xml,
-        svg
+        bpmn: this.dataXmlSvg.xml,
+        svg: this.dataXmlSvg.svg
```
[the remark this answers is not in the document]
marcoAntonioNina: Yes, I agree, but I call "modeler-save" from the package where I don't have the data.

resources/js/processes/screen-builder/screen.vue
```diff
@@ -2,41 +2,7 @@
...
-              </button>
-            </div>
-
-            <button v-if="permission.includes('export-screens')" type="button" :title="$t('Export Screen')" @click="beforeExportScreen" class="btn btn-secondary btn-sm ml-1"><i class="fas fa-file-export"></i></button>
-            <button @click="saveScreen(false)" type="button" class="btn btn-secondary btn-sm ml-1"><i class="fas fa-save"></i></button>
-          </b-col>
-
-        </b-row>
-      </b-card-header>
+      <menu-screen ref="menuScreen" :options="optionsMenu"></menu-screen>
```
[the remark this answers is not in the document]
marcoAntonioNina: was renamed to TopMenu

resources/js/processes/scripts/components/ScriptEditor.vue
```diff
@@ -1,12 +1,7 @@
 <template>
   <b-container class="h-100">
     <b-card no-body class="h-100">
-      <b-card-header class="text-right">
-        <b-button title="Save Script" @click="save" size="sm">
-          <i class="fas fa-save" />
-          {{ $t('Save') }}
-        </b-button>
-      </b-card-header>
+      <menu-script ref="menuScript" :options="optionsMenu"></menu-script>
```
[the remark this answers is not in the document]
marcoAntonioNina: was renamed to TopMenu

resources/views/processes/screens/edit.blade.php
marcoAntonioNina: yes, it's arming with a controller and passed that way.

resources/views/processes/scripts/edit.blade.php
marcoAntonioNina: yes, it's arming with a controller and passed that way.

resources/js/components/Menu.vue
marcoAntonioNina: yes,  "ref" was actually used to call these methods

resources/js/processes/modeler/components/ModelerApp.vue
```diff
@@ -100,18 +100,18 @@
...
-    actionSaveBpmn ({xml, svg}) {
+    emitRegisteredEvents ({xml, svg}) {
       this.dataXmlSvg.xml = xml;
       this.dataXmlSvg.svg = svg;
 
       this.externalEmit.forEach(item => {
         window.ProcessMaker.EventBus.$emit(item);
       })
-      if (this.saveEmitDefault) {
+      if (!this.externalEmit) {
```
chipit24, line 110: How is `externalEmit` modified? It seems it's set to an empty array and never modified. An empty array is truthy, so `!this.externalEmit` will always evaluate to `false`, so it seems this `if` block can be removed. Is that correct?

resources/js/processes/screen-builder/screen.vue
```diff
@@ -296,8 +361,41 @@ import formTypes from "./formTypes";
...
+      });
+      ProcessMaker.EventBus.$on('open-form-watchers',  () => {
+        this.openWatchersPopup();
+      });
+      ProcessMaker.EventBus.$on('export-screen',  () => {
+        this.beforeExportScreen();
+      });
+      ProcessMaker.EventBus.$on('save-screen',  (value, resolve, reject) => {
+        this.saveScreen(value, resolve, reject);
+      });
```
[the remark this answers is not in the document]
marcoAntonioNina: was changed to functions

resources/js/processes/modeler/components/ModelerApp.vue
marcoAntonioNina: Yes, this array is established by an external package, which records events.

resources/js/processes/scripts/components/ScriptEditor.vue
```diff
@@ -1,12 +1,7 @@
 <template>
   <b-container class="h-100">
     <b-card no-body class="h-100">
-      <b-card-header class="text-right">
-        <b-button title="Save Script" @click="save" size="sm">
-          <i class="fas fa-save" />
-          {{ $t('Save') }}
-        </b-button>
-      </b-card-header>
+      <top-menu ref="menuScript" :options="optionsMenu"/>
```
chipit24, line 4: This ref isn't used anywhere and can be removed.

```suggestion
      <top-menu :options="optionsMenu"/>
```

PR #3072, opened 2020-04-23
Add documentation view button for package-process-documenter

johnblackspear:
This adds a documentation link in the processes list.

The documentation button shows conditionally if the user has `view-processes` privileges and the `package-process-documenter` is installed.

## Determining if a package is installed

This PR adds a new helper file, `ProcessMaker/PackageHelper.php` that allows us to easily determine whether a package is installed (e.g. by calling `\ProcessMaker\PackageHelper::isPmPackageProcessDocumenterInstalled()`).

### Without package-process-documenter
image

### With package-process-documenter installed
image

Fixes: https://github.com/ProcessMaker/package-process-documenter/issues/12

### DIFF
--- a/resources/js/processes/components/ProcessesListing.vue
+++ b/resources/js/processes/components/ProcessesListing.vue
@@ -72,7 +72,7 @@
               </b-btn>
               <b-btn
                       variant="link"
-                      @click="onAction('action-todo', props.rowData, props.rowIndex)"
+                      @click="onAction('view-documentation', props.rowData, props.rowIndex)"
                       v-b-tooltip.hover
                       :title="$t('Documentation')"
                       v-if="permission.includes('view-processes') && isDocumenterInstalled"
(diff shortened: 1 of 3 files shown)


resources/js/processes/components/ProcessesListing.vue
```diff
@@ -70,6 +70,15 @@
               >
                 <i class="fas fa-cog fa-lg fa-fw"></i>
               </b-btn>
+              <b-btn
+                      variant="link"
+                      @click="onAction('view-documentation', props.rowData, props.rowIndex)"
+                      v-b-tooltip.hover
+                      :title="$t('Documentation')"
```
chipit24, line 77: Not sure if this was discussed elsewhere and I missed it, but it seems, according to the issue, that this should say "View Documentation".

johnblackspear: Good catch, thanks

johnblackspear:
Thanks for finding that BBox error man!

PR #3180, opened 2020-06-03
Auth Client Pagination Missing

### LINKED ISSUE #3159: Auth Client Pagination Missing
Route: /admin/auth-clients

Need to add the proper vue-table with pagination and filtering.

marcoAntonioNina:
Fixed #3159 

## Changes
- Fix paginator Auth Clients to change Per Page

### DIFF
--- a/resources/js/admin/auth-clients/components/AuthClientsListing.vue
+++ b/resources/js/admin/auth-clients/components/AuthClientsListing.vue
@@ -139,6 +139,10 @@ export default {
 
       //Pagination
       let meta = {};
+      if (this.perPage >= data.length) {
+        this.page = 1;
+      }
+
       meta.total = data.length;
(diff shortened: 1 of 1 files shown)


resources/js/admin/auth-clients/components/AuthClientsListing.vue
```diff
@@ -139,6 +139,10 @@ export default {
 
       //Pagination
       let meta = {};
+      if (this.perPage >= data.length) {
```
velkymx, line 142: Doesn't the pagination come down as part of the axios returned object? Why don't we just reference it?

image

marcoAntonioNina: the api is from laravel Passport and does not contain the required meta data.

PR #3228, opened 2020-06-24
Timeline styles

### LINKED ISSUE #3223: Request History Missing
When opening a request detail, the process history is missing from the bottom of the page. 

Ex: https://staging-pm4.processmaker.net/requests/78

image

danloa:
Resolves #3223 

Fix to style issues in the time line:
- Do not add icons to replies and reactions
- Pass explicitly the comment status to timeline items

### DIFF
--- a/resources/views/requests/show.blade.php
+++ b/resources/views/requests/show.blade.php
@@ -230,7 +230,7 @@
                               :edit="configurationComments.edit"
                               :remove="configurationComments.remove"
                               :adding="configurationComments.comments"
-                              :readonly="'request.status' === 'COMPLETED'"
+                              :readonly="request.status === 'COMPLETED'"
                               />
             </div>
             <div class="ml-md-3 mt-3 mt-md-0">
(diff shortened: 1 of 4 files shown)


resources/views/requests/show.blade.php
```diff
@@ -230,6 +230,7 @@ class="nav-link"  @click="switchTab('editdata')">
                               :edit="configurationComments.edit"
                               :remove="configurationComments.remove"
                               :adding="configurationComments.comments"
+                              :readonly="'{{$request->status}}' === 'COMPLETED'"
```
caleeli, line 233: try to use the javascript/vue variable `request` instead of blade variable, to have something like this:
`:readonly="request.status === 'COMPLETED'"`
to be reactive with request status update

resources/views/requests/show.blade.php
```diff
@@ -230,6 +230,7 @@ class="nav-link"  @click="switchTab('editdata')">
                               :edit="configurationComments.edit"
                               :remove="configurationComments.remove"
                               :adding="configurationComments.comments"
+                              :readonly="'request.status' === 'COMPLETED'"
```
caleeli, line 233: That looks wrong, you are comparing to static strings

PR #3296, opened 2020-07-28
FOUR-1557 Allow users to access nested screens if they have access to the parent

nolanpro:
Fixes https://processmaker.atlassian.net/browse/FOUR-1557

Allow users to access nested screens if they have access to the parent

By adding a new route we can avoid modifying screen builder


nolanpro:
Closing since based off wrong branch

PR #3308, opened 2020-07-31
Identity & Access Management: Session timeout

gustavo-romero:
Proposal to change the default session timeout to 15 minutes with a 90 second session expire warning as an industry best security practice.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -29,7 +29,7 @@ return [
     |
     */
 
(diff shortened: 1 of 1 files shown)


velkymx:
If you'd like to configure the session settings it should be done by the CloudOps team in the ENV file.

gustavo-romero:
I understand, but this should be the default behavior as a security best practice @velkymx
Other use cases where 15 minutes is too short, could be set manually in .env

PR #3319, opened 2020-08-04
Bugfix/bpmn queue priority

caleeli:


ProcessMaker/Console/Commands/MultipleRequest.php
```diff
@@ -0,0 +1,85 @@
+<?php
```
danloa, line 1: This file is not needed for the horizon improvement, right?

PR #3350, opened 2020-08-18
Allow to retry script tasks that are in status ACTIVE

### LINKED ISSUE #3340: Allow to retry script tasks that are in status ACTIVE
The current version of the RetryScriptTasks command runs script tasks that are in state FAILING only. In the AMS server many script tasks stalled with the status ACTIVE. 

The script must allow to run scripts in this state.

danloa:
Resolves #3340 

Added code to remove tokens from locks table before attempting run them

### DIFF
--- a/ProcessMaker/Console/Commands/RetryScriptTasks.php
+++ b/ProcessMaker/Console/Commands/RetryScriptTasks.php
@@ -50,7 +50,7 @@ class RetryScriptTasks extends Command
             exit($this->error('No failing script tasks found.'));
         }
 
-   	    $requestIds = $tasks->pluck('process_request_id');
+        $requestIds = $tasks->pluck('process_request_id');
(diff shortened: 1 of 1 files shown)


ProcessMaker/Console/Commands/RetryScriptTasks.php
```diff
@@ -45,6 +46,9 @@ public function handle()
     {
         $tasks = $this->retrieveTaskList();
 
+   	    $requestIds = $tasks->pluck('process_request_id');
+        ProcessRequestLock::whereIn('process_request_id', $requestIds)->delete();
+
         if (!$tasks->count()) {
```
velkymx, line 52: Move this count above the ProcessRequestLock. No need to run ProcessRequestLock if the task->count is zero.

PR #3356, opened 2020-08-20
Add filter data soruces

caleeli:
Add filter data sources

### DIFF
--- a/ProcessMaker/Traits/MakeHttpRequests.php
+++ b/ProcessMaker/Traits/MakeHttpRequests.php
@@ -53,6 +53,12 @@ trait MakeHttpRequests
             $url .= $separator . $config['queryString'];
         }
 
+        // If exists a filter string in the call, add it to the URL
+        if (array_key_exists('filter', $config)) {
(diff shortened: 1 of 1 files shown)


ProcessMaker/Traits/MakeHttpRequests.php
```diff
@@ -53,6 +53,12 @@ public function request(array $data = [], array $config = [])
             $url .= $separator . $config['queryString'];
         }
 
+        // If exists a filter string in the call, add it to the URL
+        if (array_key_exists('filter', $config)) {
+            $separator = strpos($url, '?') ? '&' : '?';
+            $url .= $separator . 'filter='. urlencode($config['filter']);
```
velkymx, line 59: Don't we already do this by adding a filter={filter} in the End Point URL in the Data Source?

caleeli: the default data connector created for collection does not includes `filter` parameter
That's why it was added dynamicaly like `pmql` parameter
image

PR #3359, opened 2020-08-20
Add permission to view other user's profiles

nolanpro:
Fixes https://processmaker.atlassian.net/browse/FOUR-1721

Adds a new view-other-users-profiles permission.

I also had to refactor the permissions seeder so that it can add new permissions that don't exist.

IMPORTANT: after this is deployed, devops will need to run `php artisan db:seed --class=PermissionSeeder` to add the new permission

image

### DIFF
--- a/ProcessMaker/Http/Controllers/ProfileController.php
+++ b/ProcessMaker/Http/Controllers/ProfileController.php
@@ -57,8 +57,15 @@ class ProfileController extends Controller
      *
      * @return \Illuminate\View\View|\Illuminate\Contracts\View
      */
-    public function show($id)
+    public function show(Request $request, $id)
     {
+        if (
+            $request->user()->id !== intval($id) &&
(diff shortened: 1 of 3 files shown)


ProcessMaker/Http/Controllers/ProfileController.php
```diff
@@ -57,8 +57,15 @@ function ($result, $item) {
...
      * @return \Illuminate\View\View|\Illuminate\Contracts\View
      */
-    public function show($id)
+    public function show(Request $request, $id)
     {
+        if (
+            $request->user()->id !== intval($id) &&
+            !$request->user()->can('view-other-users-profiles')
+        ){
+            abort(404);
```
caleeli, line 66: Should not be better a 401 status code?

nolanpro: The issue specifically says 404

PR #3438, opened 2020-10-26
Remove magic variables only when the process has completed

nolanpro:
https://processmaker.atlassian.net/browse/FOUR-2146

Removes `_user`, `_request`, `_parent` from the request data when the process completes instead of after each completed task


velkymx:
@nolanpro this doesn't satisfy the ticket. 

Expected Behavior: The `_user` variable never disappears. 

Note this screenshot: 

image

See how in the sidebar the user is assigned? Now when you look at the data object the `_user` variable is missing.

nolanpro:
@velkymx which server are you testing this on? It looks like the PR did not auto deploy to the server. I can work with dev ops to figure out why that happened

PR #3453, opened 2020-11-04
Changed setting order and label names.

PMQABen:
Fixes https://processmaker.atlassian.net/browse/FOUR-2191

### DIFF
--- a/resources/js/processes/screen-builder/components/file-upload-control.js
+++ b/resources/js/processes/screen-builder/components/file-upload-control.js
@@ -17,10 +17,20 @@ export default {
     inspector: [
       {
         type: "FormInput",
+        field: "name",
+        config: {
+          label: "Variable Name",
+          name: 'Name',
+          helper: "A variable name is a symbolic name to reference information",
+          validation: 'regex:/^(?:[A-Z_.a-z])(?:[0-9A-Z_. \/a-z])*$/|required'
+        }
+      },
+      {
+        type: "FormInput",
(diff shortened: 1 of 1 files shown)


resources/js/processes/screen-builder/components/file-upload-control.js
```diff
@@ -17,28 +17,28 @@ export default {
...
       },
       {
         type: "FormInput",
-        field: "name",
+        field: "accept",
         config: {
-          label: "Name",
-          name: 'Name',
-          helper: "The name of the upload",
-          validation: 'regex:/^(?:[A-Z_.a-z])(?:[0-9A-Z_. \/a-z])*$/|required'
```
velkymx, line 41: I think we should keep the validation...

velkymx (changes requested):
I think we should keep the validation. Having examples of the mime-times is a good step though!!!

resources/js/processes/screen-builder/components/file-upload-control.js
PMQABen: The validation for "Name" is being kept, we're renaming it to "Variable Name" and moving it to the top to match how the rest of the screen builder configs are laid out.

PMQABen: image

velkymx: Great

PR #3492, opened 2020-11-23
Fix anonymous user showing in add group member users select list

nolanpro:
Fixes https://processmaker.atlassian.net/browse/FOUR-2333

Add nonSystem filter to group members controller


caleeli (approved):
Just good

PR #3508, opened 2020-11-30
Fix passport dependency

danloa:
Fixes dependency problem with laravel passport


velkymx:
Not needed for the 4.1 branch

PR #3525, opened 2020-12-03
FOUR-2416: Advanced screen does not display any content even by using the default code

danloa:
Resolves [https://processmaker.atlassian.net/browse/FOUR-2416](https://processmaker.atlassian.net/browse/FOUR-2416)

Code that allows other renderers (like advanced forms) has been added.


caleeli (changes requested):
Please  move to the screen package

danloa:
The solution will be implemented in screen builder so this PR is not necessary

PR #3526, opened 2020-12-03
FOUR-2444: When running scripts tasks an error is generated

danloa:
Resolves [https://processmaker.atlassian.net/browse/FOUR-2444](https://processmaker.atlassian.net/browse/FOUR-2444)

A similar problem was found in PM 4.0.15. A dependency of Laravel passport has been updated the last week and is generating problems. The solution is to fix the library

"lcobucci/jwt": "3.3.1"

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "3a1b89cca74cafd7882577b65e906744",
+    "content-hash": "d62b192021456e388e7932adffa26971",
     "packages": [
         {
             "name": "babenkoivan/elastic-adapter",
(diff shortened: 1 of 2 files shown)


composer.lock
```diff
@@ -5404,25 +5795,24 @@
...
-                "url": "https://api.github.com/repos/spatie/pdf-to-image/zipball/6718a89065c34dfadfbb2ca61eb1c3e6a956e503",
-                "reference": "6718a89065c34dfadfbb2ca61eb1c3e6a956e503",
+                "url": "https://api.github.com/repos/spatie/pdf-to-image/zipball/ea81a2d5f4fb0ca9f4c4fc0d4f8813135e993b2d",
+                "reference": "ea81a2d5f4fb0ca9f4c4fc0d4f8813135e993b2d",
                 "shasum": ""
             },
             "require": {
-                "org_heigl/ghostscript": "^2.2",
-                "php": "^7.0"
+                "ext-imagick": "*",
```
ryancooley, line 5811: This appears to add a requirement for Image Magick. Do we know why and do we know if our servers are prepared to handle that requirement?

danloa: 1) The latest version of pdf-to-image needs that extension.  And this package is a requirement of laravel/media-library... It seems we can't avoid this requirement.
2) Many of the updates are additions of support and funding information (those are added when using composer 2.0+. Composer 1 doesn't add them.

ryancooley (approved):

boliviacoca (approved):

PR #3569, opened 2020-12-11
Fix signal count on empty collection

caleeli:
Fix signal count on empty collection


pmkerberos:
QA server was successfully deployed https://bugfix-signal_count-processmaker.ci.processmaker.net

boliviacoca (approved):

PR #3688, opened 2021-02-24
Fix issue with fallback routing in packages

ryancooley:
## Changes
- Fixes an issue where a 500 error was returned instead of a 404 error in certain package API routes (for example: Collections and Saved Search)
  - Refactors the way a fallback route is determined when a model is not found in a bound route
    - Many packages do not namespace API route names with the `api.` prefix but rather with the name of the package
    - This caused the fallback routing to return the web fallback route when a model could not be found, which in turn generated a 500 error because web middleware was being used on an API route
    - Now we determine whether a fallback route exists under the package namespace
      - If so, we return it
      - If not, we check to see whether API middleware is in use and fallback to our API fallback route if so

Closes https://processmaker.atlassian.net/browse/FOUR-2882

### DIFF
--- a/ProcessMaker/Exception/Handler.php
+++ b/ProcessMaker/Exception/Handler.php
@@ -85,7 +85,7 @@ class Handler extends ExceptionHandler
         $prefixes = [];
         $prefixes[] = explode('.', $route->getName())[0];
         if (isset($route->computedMiddleware) && count($route->computedMiddleware)) {
-            if (in_array('api', $route->computedMiddleware) || in_array('aduth:api', $route->computedMiddleware)) {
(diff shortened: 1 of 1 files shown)


pmkerberos:
QA server was successfully deployed https://bugfix-four-2882-processmaker.ci.processmaker.net

pmkerberos:
QA server was successfully deployed https://bugfix-four-2882-processmaker.ci.processmaker.net

nolanpro (approved):

boliviacoca (approved):

pmkerberos:
QA server was successfully deployed https://develop-processmaker.ci.processmaker.net

PR #3706, opened 2021-03-10
Auth Package support

ryancooley:
## Changes
- Adds items needed to support the new Authentication package
  - Adds a LoginManager which allows for addons to the login form
  - Adds a link to the ProcessMaker SocialiteProviders repo to composer.json, which is needed to properly install the Auth package


composer.json
```diff
@@ -97,6 +97,12 @@
       "@php artisan package:discover"
     ]
   },
+  "repositories": [
```
caleeli, line 100: This repo was added, but there is no any dependency added, is that right?

ProcessMaker/Managers/LoginManager.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * List addons
+     *
+     */
+    public function list()
+    {
+        return $this->addons;
+    }
+
+}
```
caleeli, line 68: Please add a newline at the end of the file

ryancooley: Done!

caleeli (approved):

boliviacoca (approved):

PR #3729, opened 2021-03-22
Four 3015

PMQABen:
Fixed the formatting to match what was actually requested, and then moved it up in the file to better match the select drop-down format for editing user profiles.

So let's try this again.


marcoAntonioNina:
@PMQABen the change looks good, but when testing it shows something weird.
image

also the correct branch is 4.1.1 and not develop.

PMQABen:
I didn't run into that weird i issue on my end, but I'll close this pull and work on it further.

PR #3733, opened 2021-03-25
Restrict users without "view all requests" permission to participated-in requests

ryancooley:
## Changes
- Adds a restriction to the Requests index endpoint to prevent users without "view all requests" permission from seeing requests in which they have not participated

https://processmaker.atlassian.net/browse/FOUR-2309


ryancooley:
Invalid

PR #3756, opened 2021-04-12
Ethos Events Package requirements

gusys:
The conditionalStartEvent and Process elements are exposed for the will extend InspectorsExtension for Ellucian Ethos Events Package


gusys:
This pull request is no longer necessary, the Ellucian Ethos Events package will extend Inspector of  the StartEvent element

PR #3863, opened 2021-06-29
FOUR-3612: Review and improve sql query request_user_permissions

danloa:
Fixes [https://processmaker.atlassian.net/browse/FOUR-3612](https://processmaker.atlassian.net/browse/FOUR-3612)

The query was modified to use a join and the user_id index to avoid the subquery that usually is slower
Before the modification the MySql planner didn't use the user_index:

image

After the modification the index is used:
image


caleeli (approved):
image

very good improvement

boliviacoca (approved):

PR #4078, opened 2021-10-12
Fix the start permissions of a user nested in a group

sanjacornelius:
Ticket: [1056](http://tickets.pm4overflow.com/tickets/1056)

<h2>Changes</h2>

- Update the Process Policy to enable start permissions for users in a group that is nested.

<h2>To Test</h2>

- Create Groups **"Group1"** and **"Group2"**
- Create User **"User2"**
- Assign **"Group2"** as a member of **"Group1"**
- Assign **"User2"** to **"Group2"**
- Create a process **"Test Process**" and configure its Start Event to be started by **"Group1"**
- Login as **"User2"**
- Try to Start **"Test Process"**

**Expected Behavior**

The User assigned as request starters via a group of groups should be able to start a request.

### DIFF
--- a/ProcessMaker/Policies/ProcessPolicy.php
+++ b/ProcessMaker/Policies/ProcessPolicy.php
@@ -7,6 +7,7 @@ use ProcessMaker\Models\Process;
 use Illuminate\Auth\Access\HandlesAuthorization;
 use Illuminate\Http\Request;
 use ProcessMaker\Models\AnonymousUser;
+use ProcessMaker\Models\GroupMember;
 
 class ProcessPolicy
 {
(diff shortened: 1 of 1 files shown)


ProcessMaker/Policies/ProcessPolicy.php
```diff
@@ -35,8 +36,10 @@ public function before(User $user)
      */
     public function start(User $user, Process $process)
     {
-        $groupIds = $user->groups->pluck('id');
-
+        $userGroupIds = $user->groups->pluck('id')->all();
+        $nestedGroupIds = GroupMember::whereIn('member_id', $userGroupIds)->pluck('group_id')->all();
```
ryancooley, line 40: I think we need to add another `where` here to ensure that the `member_type` is `ProcessMaker\Models\Group` otherwise we may end up getting user IDs in this query as well.

PR #4079, opened 2021-10-12
Catch conditional errors to do not break to others

caleeli:
Fixes: https://processmaker.atlassian.net/browse/FOUR-3681
When a StartConditionalEvent has an invalid expression, it throws and exception and breaks, the evaluations of other events like the Timer StartEvents.

This Catch conditional errors to do not break to others.


pmkerberos:
QA server was successfully deployed https://bugfix-four-3681-processmaker.ci.processmaker.net

PR #4106, opened 2021-10-20
Task/ticket-1154: Changed avatar class color from warning to info

agustinbusso:
Ticket [1154](http://tickets.pm4overflow.com/tickets/1154)
Changed avatar color to info

### DIFF
--- a/resources/js/components/AvatarImage.vue
+++ b/resources/js/components/AvatarImage.vue
@@ -121,14 +121,14 @@ export default {
           return true;
         }
       };
-      
+
       return false;
     },
     variant(value) {
       if (value.src) {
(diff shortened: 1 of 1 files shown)


resources/js/components/AvatarImage.vue
```diff
@@ -121,14 +121,14 @@ export default {
...
-      
+
       return false;
     },
     variant(value) {
       if (value.src) {
         return 'secondary';
       } else {
-        return 'warning';
+        return 'info';
```
josechirivella, line 131: nit comment. Another way to do it...
```javascript
if (value.src) return 'secondary';
return 'info';
```

PR #4137, opened 2021-11-10
Fix 401s redirecting to home page for data source tests

nolanpro:
Part of https://processmaker.atlassian.net/browse/FOUR-4317

When we get 401s from any api call it automatically redirects the user to the home page. Data sources can return 401s that we do not want to redirect. There was already an exception if the URL contains 'request'. This fix adds one for the 'test' endpoint.


pmkerberos:
QA server was successfully deployed https://bugfix-four-4317-processmaker.ci.processmaker.net

marcoAntonioNina (approved):

Issue #4161, opened 2021-11-18
Add build number to about page


ryancooley:
**Is your feature request related to a problem? Please describe.**
As a developer I'd like to know the exact commit a version of ProcessMaker is running on.

**Describe the solution you'd like**
Add a build number underneath the copyright notice. Use the `<small>` tag to make the text smaller.

<img width="749" alt="Screen Shot 2021-11-18 at 11 19 22 AM" src="https://user-images.githubusercontent.com/867714/142482505-391423e9-3249-4fd6-bbcc-7840b1bef35a.png">

runyan-co:
Ready for review @ryancooley.

PR #4162, opened 2021-11-18
[4.1][Feature] Show commit hash on about page

runyan-co:
## Issue & Reproduction Steps
Developer's want to know the exact commit hash for a given build without having to necessarily have command line access. This will enable better debugging for builds deployed to production and other environments.

## Solution
We've added a "processmaker" property to the "extra" property in composer.json. Within that, there is a "build" property which contains a string value representing the commit hash of the build. This is read by the AboutController and shown on the about page (/about) in the footer. This value will be set automatically when builds are released.

If this value isn't present, the exception will be caught and the build commit hash won't be attempted to be shown.

## How to Test
Check the bottom of the /about page and there should be small text under the copyright notice with the commit hash info for the build. If you remove the "processmaker" attribute from the composer.json "extra" attribute, the commit hash info should disappear from the page without any errors or uncaught exceptions.

## Related Tickets & Packages
Related [GitHub issue](https://github.com/ProcessMaker/processmaker/issues/4161).

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Http/Controllers/AboutController.php
+++ b/ProcessMaker/Http/Controllers/AboutController.php
@@ -2,7 +2,9 @@
 
 namespace ProcessMaker\Http\Controllers;
 
+use Exception;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
 use ProcessMaker\Http\Controllers\Controller;
 use ProcessMaker\Models\Setting;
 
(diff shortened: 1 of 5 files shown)


pmkerberos:
QA server was successfully deployed https://feature-include-commit-hash-on-about-page-processmaker.ci.processmaker.net

ryancooley (approved):

PR #4173, opened 2021-11-22
Fixing issue with cache get all into cache_settings when force args is enabled

epdl:
## Issue & Reproduction Steps
When cache_settings is using a force=true argument and the key 'all' is not defined, the function reset resttings but don't initialize the $cache again.
Issue can be reproduced running php `artisan package-ellucian-ethos:install`

## Solution
- Adding the $cache init command after the reset

## How to Test
Execute `artisan package-ellucian-ethos:install` without errors 

## Related Tickets & Packages
- github.com/ProcessMaker/package-ellucian-ethos [private]

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Helpers/SettingsHelper.php
+++ b/ProcessMaker/Helpers/SettingsHelper.php
@@ -94,6 +94,7 @@ if (!function_exists('cache_settings')) {
                 // are available if they aren't cached yet
                 if (!$cache->has('all')) {
                     $settings = settings();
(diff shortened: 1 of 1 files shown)


ryancooley (approved):

runyan-co (approved):
I wasn't able to reproduce this locally using the package-ellucian-ethos install command, but  since this is verifiably happening in production and the fix is very simple, I'll approve and merge.

PR #4192, opened 2021-12-02
Redirect interstitial tasks from sub processes

nolanpro:
See https://github.com/ProcessMaker/screen-builder/pull/1118

### DIFF
--- a/ProcessMaker/Http/Controllers/RequestController.php
+++ b/ProcessMaker/Http/Controllers/RequestController.php
@@ -127,27 +127,6 @@ class RequestController extends Controller
         ));
     }
 
-    /**
-     * Redirect to owner task of a subprocess
-     *
-     * @param ProcessRequest $request
-     *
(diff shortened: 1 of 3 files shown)


pmkerberos:
QA server was successfully deployed https://bugfix-four-4536-processmaker.ci.processmaker.net

caleeli:
The issue was caused by a missing deploy of the screen-builder change. Fixed in test environment.

The following processes were tested:

[x] A sub processes with interstitial tasks
image
[x] Using intermediate script tasks before and after each interstitial task
image
[x] With Assignment by Value
[x] Tested with exclusive gateways and a loop back flow
[x] WebEntry + Interstitial + CallActivity
[x] Two different requests run in parallel with the same user
[x] Same request running in parallel with the same user in two different computers (they run in sync)

Limitations:
- Only two levels of Sub Processes are supported (Parent -> Child -> Parent)
- Unable to use a sub process immediately after Web Entry

caleeli (approved):

PR #4193, opened 2021-12-03
[4.2] Multiple copies of the process are created and import doesn't finish

runyan-co:
## Issue & Reproduction Steps
When attempting to import a process from a file, the import failed and the user was left with the "spinning" import screen without any update and with the process imported three times. This happened because the ImportProcess job ran three times and even though it imported the process, it was failing when attempting to import assets referenced in the BPMN which didn't exist in the import file data itself.

To reproduce the issue:

1. Import the [attached processes](https://github.com/ProcessMaker/processmaker/files/7645849/process-imports.zip) on a 4.2.* instance.
2. See that the process import will not finish and remains "spinning".
3. Navigate back to the processes list
4. Notice there are now 3 incomplete copies of each process after attempting to the import.

## Solution
The ImportProcess job will only be attempted one time in case of a failure (to prevent duplicates). The primary solution was adding a validation to ensure the array of a screens actually contained the references screen id.

## How to Test
You should be able to repeat the steps above without the ImportProcess job failing and without being left on a "spinning"/waiting screen for the import to finish.

## Related Tickets & Packages
[Related JIRA FOUR-4469 ticket](https://processmaker.atlassian.net/browse/FOUR-4469).
[Related PR for package-webentry](https://github.com/ProcessMaker/package-webentry/pull/132).

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Assets/ScreensInProcess.php
+++ b/ProcessMaker/Assets/ScreensInProcess.php
@@ -69,8 +69,12 @@ class ScreensInProcess
         $nodes = $xpath->query("//*[@pm:screenRef!='']");
         foreach ($nodes as $node) {
             $oldRef = $node->getAttributeNS(WorkflowServiceProvider::PROCESS_MAKER_NS, 'screenRef');
-            $newRef = $references[Screen::class][$oldRef]->getKey();
-            $node->setAttributeNS(WorkflowServiceProvider::PROCESS_MAKER_NS, 'screenRef', $newRef);
+            if (array_key_exists($oldRef, $references[Screen::class])) {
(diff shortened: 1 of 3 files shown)


pmkerberos:
QA server was successfully deployed https://bugfix-four-4469-processmaker.ci.processmaker.net

caleeli (changes requested):
@runyan-co, It looks like it is still falling for this process:

[Admissions Application.json.txt](https://github.com/ProcessMaker/processmaker/files/7661199/Admissions.Application.json.txt)

https://user-images.githubusercontent.com/8028650/144859481-fa6340d5-97c2-472f-abfd-5b9be9efd306.mp4

caleeli (approved):

PR #4213, opened 2021-12-15
4.2 Fix Filtering request by User when username contains numbers 

sanjacornelius:
## Issue & Reproduction Steps
Ticket: [FOUR-4910](https://processmaker.atlassian.net/browse/FOUR-4910)

When doing a PMQL query we are passing the value in a try-catch block to see if the value is a parsable date if not then we return the original string. Values with numbers such as `W0584` are being detected as a parsable date, which is causing this error. 

## Solution

- Standardize a date format in the PMQL query based on the [PMQL Date Comparison documentation ](https://github.com/ProcessMaker/pmql/blob/master/README.md#date-comparison) and check if the value is in this date format before parsing. 

## Note

This does change how users would query for dates. Users now need to properly format the date in the query `yyyy-mm-dd` example query `Started > "1990-01-01"`

## How to Test

1. Create a user that includes numbers in the username (i.e W0584 or 0584123), the rest of the fields does not matter
2. Go to requests, try to filter by the User field, and select the recently created user

**Expected Behavior**
No errors are returned and if data is available that data is listed.

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Traits/ExtendedPMQL.php
+++ b/ProcessMaker/Traits/ExtendedPMQL.php
@@ -143,7 +143,7 @@ trait ExtendedPMQL
                 default: 
                     // Check to see if the value is a date/datetime formatted if not return original value
                     $isDateFormatted = Carbon::hasFormatWithModifiers($value, 'Y#m#d');
-                    $isDateTimeFormatted = Carbon::hasFormatWithModifiers($value, 'Y#m#d h:i:s');
+                    $isDateTimeFormatted = Carbon::hasFormatWithModifiers($value, 'Y#m#d H:i:s');
                     if ($isDateFormatted || $isDateTimeFormatted) {
                         $value = $this->parseDate($value);
                     }
--- a/ProcessMaker/Traits/ExtendedPMQL.php
+++ b/ProcessMaker/Traits/ExtendedPMQL.php
@@ -154,13 +154,18 @@ trait ExtendedPMQL
     }
 
     private function parseDate($value) {
-        $parsed = Carbon::parse($value, auth()->user()->timezone);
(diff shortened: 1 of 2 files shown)


pmkerberos:
QA server was successfully deployed https://bugfix-four-4904-4-2-processmaker.ci.processmaker.net

ProcessMaker/Traits/ExtendedPMQL.php
```diff
@@ -135,20 +135,32 @@ private function parseValue($expression)
...
-                    $value = $parsed->toDateTimeString();
+        if ((is_string($value) && strlen($value) > 1)) {
+            switch ($value) {
+                case $value instanceof IntervalExpression: 
+                    $value = $this->parseDate($value);
+                    break;
+                default: 
+                    // Check to see if the value is a date/datetime formatted if not return original value
+                    $isDateFormatted = Carbon::hasFormatWithModifiers($value, 'Y#m#d');
+                    $isDateTimeFormatted = Carbon::hasFormatWithModifiers($value, 'Y#m#d h:i:s');
```
nolanpro, line 146: @sanjacornelius I think we should use `H` here instead of `h` since we're using a 24 hour clock in other places and the format doesn't specify AM or PM

nolanpro (changes requested):

ProcessMaker/Traits/ExtendedPMQL.php
sanjacornelius: @nolanpro  Good catch! Code has been updated thank you!

pmkerberos:
QA server was successfully deployed https://bugfix-four-4904-4-2-processmaker.ci.processmaker.net

nolanpro (approved):

PR #4240, opened 2022-01-11
FOUR-4962: Ethos User Sync Issues in Ellucian 4.2.x

danloa:
## Issue & Reproduction Steps
Some errors are not handled correctly by PHP. If this happens when running script tasks they can be left in status ACTIVE without executing the job.

Some options to simulate this condition are:

1. To simulate an unhandled error in development mode,  modify the  file ProcessMaker/Managers/WorkflowManager.php and add the next line in the runScriptTask method:

Illuminate\Support\Facades\Log::Info( print_r($token, true));

As $token has circular references  it will generate an exhaust memory fatal error.

2. Another way to simulate the condition of  a halted script task is to run a simple process that contains a script task (for example start event->script task-> end event) . After that the  end event token (in table process_reuest_tokens) shoul be deleted and the script task token set its status to "ACTIVE"

## Solution
- A new command processmaker:garbage-collect has been created. This command can be configured to be executed in the server's crontab periodically to handle all  halted sctipt tasks.
- The BpmnSubscriber class now hast a function registerErrorHandler that registers a PHP's shut_down_function that is executed when PHP encounters and unhandled error that forces a PHP process shutdown. This happens when the memory is exhausted, for example (this condition can be reproduce with the option 1, described above)

## How to Test
 - Use one of the 2 ways to reproduce the error.
 - Execute from the command line: php artisan processmaker:garbage-collect
 - Verify that the halted script tasks were executed and the processes were unblocked.

## Related Tickets & Packages
- Link to any related FOUR tickets:  [https://processmaker.atlassian.net/browse/FOUR-4962](https://processmaker.atlassian.net/browse/FOUR-4962)

## Code Review Checklist


pmkerberos:
QA server was successfully deployed https://bugfix-four-4962-processmaker.ci.processmaker.net

ProcessMaker/Listeners/BpmnSubscriber.php
```diff
@@ -40,6 +40,40 @@
...
+     * @param $element
+     * @param TokenInterface|null $token
+     * @return $this
+     */
+    public function registerErrorHandler($element, TokenInterface $token = null)
+    {
+        // This storage is freed on error (case of allowed memory exhausted)
+        $this->memory = str_repeat('*', 1024 * 1024);
+
+        $path = getcwd();
```
caleeli, line 55: Could you move the storage path, like: `storage_path("logs/unhandled_error.log")`

caleeli (changes requested):
Do not use the current working directory to store files, use the storage path instead

ProcessMaker/Console/Commands/GarbageCollector.php
```diff
@@ -0,0 +1,119 @@
...
+            if ($token->element_type === 'serviceTask') {
+                RunServiceTask::dispatch($process, $instance, $token, []);
+            }
+        }
+        $bar->finish();
+    }
+
+    private function processUnhandledErrors()
+    {
+        $fileName = getcwd(). '/unhandled_error.txt';
```
caleeli, line 86: Use storage folder instead current working directory to store files

pmkerberos:
QA server was successfully deployed https://bugfix-four-4962-processmaker.ci.processmaker.net

ryancooley (approved):

PR #4258, opened 2022-01-25
Feature/four-5129: upgrade command (4.2)

runyan-co:
## Issue & Reproduction Steps
As part of the 4.2 settings change we need an easy way to upgrade from the 4.1 .env settings to the new settings UI.  

Requirements:
- Should only allow for one run.
- Multiple runs should fail. 
- Bonus points if we can reuse this pattern for future updates that will require commands. Ideally something similar to the php artisan migrate command, but for whatever migration of code/database/etc needed

## Solution
- Two new commands have been added: upgrade:create and upgrade. These commands use the standard migration functionality of laravel but all of the upgrade migration files are stored in a directory called upgrades/ in the project root. 
- All of the upgrade logic should go in the up() method of the migration and the rollback logic (if needed/possible) should go in the down() method. 
- When ready, you can initiate any available upgrades by running `php artisan upgrade` which will look in the upgrades/ directory for upgrade migrations to run. Once run, laravel will make a record of it in the standard migrations table to prevent them from being run again. 
- Packages can publish their own upgrade migration files to the upgrades/ directory as well.

## Related Tickets & Packages
- [Ticket FOUR-5129](https://processmaker.atlassian.net/browse/FOUR-5129)

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Console/Commands/CreateUpgradeMigration.php
+++ b/ProcessMaker/Console/Commands/CreateUpgradeMigration.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace ProcessMaker\Console\Commands;
+
+use Illuminate\Console\Command;
+
+class CreateUpgradeMigration extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'upgrade:create {name}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Create an upgrade migration';
+
+    /**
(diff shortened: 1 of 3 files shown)


pmkerberos:
QA server was successfully deployed https://feature-four-5129-4-2-processmaker.ci.processmaker.net

pmkerberos:
QA server was successfully deployed https://feature-four-5129-4-2-processmaker.ci.processmaker.net

agustinbusso (approved):

ryancooley (approved):

PR #4265, opened 2022-01-28
FOUR-5118: Issue - Save search - CreditAccess dev server (for 4.1.20)

danloa:
## Issue & Reproduction Steps
When there are 1000+ requests and in every one of the the "data" json field contains a very large attribute, the endpoint requests returns a 500 error. This is noticeable when accessing the site remotely. In local dev. environments the problem is not as critical as in the remote access.  Saved search uses the controller of this endpoint (/requests)

## Solution
- To avoid that Laravel parses the larga data column of tokens and requests, the queries used to filter by request starter and participant were modified to get just ids.

## How to Test
Verify that the list requests (endpoint /requests) behavior has not changed. 

## Related Tickets & Packages
[https://processmaker.atlassian.net/browse/FOUR-5118](https://processmaker.atlassian.net/browse/FOUR-5118)

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Models/ProcessRequest.php
+++ b/ProcessMaker/Models/ProcessRequest.php
@@ -632,7 +632,11 @@ class ProcessRequest extends Model implements ExecutionInstanceInterface, HasMed
         $user = User::where('username', $value)->get()->first();
 
         if ($user) {
-            $requests = ProcessRequest::where('user_id', $expression->operator, $user->id)->get();
+            $requests = ProcessRequest::select('id')
+                ->where('user_id', $expression->operator, $user->id)
+                ->distinct()
+                ->get();
+
             return function ($query) use ($requests) {
                 $query->whereIn('id', $requests->pluck('id'));
             };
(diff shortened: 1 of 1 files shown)


pmkerberos:
QA server was successfully deployed https://bugfix-four-5118-4-1-20-processmaker.ci.processmaker.net

ProcessMaker/Models/ProcessRequest.php
```diff
@@ -632,7 +632,11 @@ private function valueAliasRequester($value, $expression)
         $user = User::where('username', $value)->get()->first();
 
         if ($user) {
-            $requests = ProcessRequest::where('user_id', $expression->operator, $user->id)->get();
+            $requests = ProcessRequest::select('id')
+                ->where('user_id', $expression->operator, $user->id)
+                ->distinct()
```
caleeli, line 637: Since `id` is the primary key of `process_requests`, `distinct()` should not be required, the generated query is:
```
"select distinct `id` from `process_requests` where `user_id` = ?"
```

caleeli (changes requested):
I think `distinct()` should not be used in 
```
select distinct `id` from `process_requests` where `user_id` = ?
```

pmkerberos:
QA server was successfully deployed https://bugfix-four-5118-4-1-20-processmaker.ci.processmaker.net

caleeli (approved):

PR #4277, opened 2022-02-07
Fix datasource response

caleeli:
## Issue & Reproduction Steps
201 response not consumed by Data Connectors.

## Solution
- Parse content if exists for status > 200 and <300

## How to Test

1. Create a DS
2. Call to a random endpoint that returns a response but with status 201
3. The response should by parsed

## Related Tickets & Packages
- [201 response not consumed by Data Connectors](https://processmaker.atlassian.net/browse/FOUR-5298)
- Requires: https://github.com/ProcessMaker/package-data-sources/pull/217

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Traits/MakeHttpRequests.php
+++ b/ProcessMaker/Traits/MakeHttpRequests.php
@@ -339,7 +339,7 @@ trait MakeHttpRequests
                 $content = json_decode($bodyContent, true);
                 break;
             case $status > 200 && $status < 300:
-                $content = [];
+                $content = !empty($bodyContent) ? json_decode($bodyContent, true) : [];
                 break;
             default:
                 throw new HttpResponseException($response);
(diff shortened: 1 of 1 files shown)


pmkerberos:
QA server was successfully deployed https://bugfix-four-5298-processmaker.ci.processmaker.net

ProcessMaker/Traits/MakeHttpRequests.php
```diff
@@ -303,7 +303,7 @@ private function response($response, array $data = [], array $config = [])
                 $content = json_decode($bodyContent, true);
                 break;
             case $status > 200 && $status < 300:
-                $content = [];
+                $content = empty($bodyContent) ? json_decode($bodyContent, true) : [];
```
danloa, line 306: The goal is do not clear the content if the status is > 201, right? If that is the case, the condition of this line should be reversed:     $content = empty($bodyContent) ? [] : json_decode($bodyContent, true);

Additionally, this line should be present in the method "responseWithHeaderData" too

danloa (changes requested):

ProcessMaker/Traits/MakeHttpRequests.php
caleeli: Right, fixed

danloa (approved):

ryancooley (approved):

PR #4376, opened 2022-05-11
Bugfix FOUR-6181 - error "Array to string conversion" with imported process

eiresendez:
## Issue & Reproduction Steps
When Importing a process and setting the Manager, it's saved in the database as a full Object instead of just an id. 

1. Import the attached process (notification-task.json). 
2. Assign a Process Manager.
image
3. Save
4. Run request and submit
5. At this point you will be able to replicate the problem, because the `manager_id` property was saved as an Object instead of an int.

## Solution
- Added the same function as the edit view to make sure that the ID is sent to the back-end instead of an object.

## How to Test
- Follow the reproduction steps of above. 
- When importing the process and assigning a Process Manager it will store the manager_id correctly on DB. 

## Related Tickets & Packages
- [FOUR-6181](https://processmaker.atlassian.net/browse/FOUR-6181)

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Models/ProcessRequestToken.php
+++ b/ProcessMaker/Models/ProcessRequestToken.php
@@ -246,7 +246,7 @@ class ProcessRequestToken extends Model implements TokenInterface
                 break;
             case 'manager':
                 $process = $this->process()->first();
-                return collect([$process->properties['manager_id'] ?? null]);
+                return collect([optional($process)->manager_id]);
                 break;
             default:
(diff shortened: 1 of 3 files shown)


josechirivella (approved):

agustinbusso (approved):
No error shows after import process with process manager

sanjacornelius (approved):

ryancooley (approved):

PR #4384, opened 2022-05-19
Include ENV variables for configure queue manager

caleeli:
## Solution
Add a ENV variable to configure the number of processes in Horizon config

## How to Test
Install or deploy  PM4, and use the env variables to setup the horizon parameters.
Start o Restart the horizon manager to load/reload the settings.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-6283

## Code Review Checklist


epdl (approved):
Looks good to me

config/horizon.php
```diff
@@ -145,59 +145,83 @@
             'supervisor-bpmn' => [
                 'connection' => 'redis',
                 'queue' => ['bpmn'],
-                'balance' => 'simple',
-                'processes' => 16,
-                'tries' => 1,
-                'timeout' => 3600,
+                'balance' => env('PM4_HORIZON_SUPERVISOR_BPMN_BALANCE', 'simple'),
+                'processes' => env('PM4_HORIZON_SUPERVISOR_BPMN_PROCESSES', 4),
```
epdl, line 149: Reviewing the documentation, looks like 'processes' are not compatible with 'minProcesses' and 'maxProcesses', we need to pick one only.

epdl (changes requested):

config/horizon.php
nolanpro: @epdl I think this is based on the `balance` setting. If balance is simple or false, the `processes` setting is used. If balance is `auto`, then `processes` is ignored and the `minProcesses` and `maxProcesses` is used.

nolanpro (changes requested):
@caleeli 
- `balanceMaxShift` and `balanceCooldown` are not supported until laravel 8 so we can remove those
- No matter what the balance setting is, When the `max_processes` key exists in the config, it will overwrite `processes` (see https://github.com/laravel/horizon/blob/3.0/src/ProvisioningPlan.php#L155) So we can get rid of `processes` and update the max_processes description in .env.md

nolanpro (approved):

ryancooley (approved):

PR #4396, opened 2022-05-31
Four-6337 Configure user password

caleeli:
## Contains:
Implement Soap base contract and classes to configure a User Password authentication

image

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-6337

## Code Review Checklist

### DIFF
--- a/tests/Feature/WebServiceSoapServiceCallerTest.php
+++ b/tests/Feature/WebServiceSoapServiceCallerTest.php
@@ -18,9 +18,8 @@ class WebServiceSoapServiceCallerTest extends TestCase
      */
     private $manager;
 
-    protected function setUp(): void
+    protected function setUpManager(): void
     {
-        parent::setUp();
         $this->manager = new WebServiceSoapServiceCaller;
         $this->app = app();
     }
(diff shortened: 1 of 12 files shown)


tests/Feature/WebServiceSoapServiceCallerTest.php
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @var WebServiceSoapServiceCaller
+     */
+    private $manager;
+
+    protected function setUp(): void
+    {
+        parent::__construct();
+        $this->manager = new WebServiceSoapServiceCaller;
+        $this->app = app();
```
caleeli, line 25: @nolanpro I have to add this line to have access to the `app` object. Which is required to use helpers like: `$this->instance` and `$this->mock`

tests/Feature/WebServiceSoapServiceCallerTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     *
+     * @var WebServiceSoapServiceCaller
+     */
+    private $manager;
+
+    protected function setUp(): void
+    {
+        parent::__construct();
```
nolanpro, line 23: @caleeli This should be `parent::setUp()`

Or even better, use you can use `setUpXXX` like

```
    protected function setUpManager()
    {
        $this->manager = new WebServiceSoapServiceCaller;
    }
```

caleeli: ohh! thanks! fixed

PR #4417, opened 2022-06-16
Bugfix FOUR-6006 - Warnings are displayed when running scripts breaking script execution

eiresendez:
## Issue & Reproduction Steps

Warnings are displayed when running scripts breaking script execution

1. Create a new script
2. Use the following code:

```php
$apiInstance = $api->tasks();
$result = $apiInstance->getTasks();
return $result;
```

3. Run the Script

## Solution
- If the type is not defined it is set to "string"  by default. That causes settyype problems. The solution is to explicitly set the type of the property.

## How to Test
- Rebuild the PHP Executor at `/admin/script-executors`.
- Follow the steps of above.

## Related Tickets & Packages
- [FOUR-6006](https://processmaker.atlassian.net/browse/FOUR-6006)

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Models/ProcessRequestToken.php
+++ b/ProcessMaker/Models/ProcessRequestToken.php
@@ -47,7 +47,7 @@ use Throwable;
  *   schema="processRequestTokenEditable",
  *   @OA\Property(property="user_id", type="string", format="id"),
  *   @OA\Property(property="status", type="string"),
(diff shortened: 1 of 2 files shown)


mcraeteisha (approved):
Approved!

ryancooley (approved):

PR #4435, opened 2022-07-07
FOUR-6484 & FOUR-6485: patch-1-cartegraph webentry fixes

runyan-co:
## Environment Setup
- Set core to this PR's branch, bugfix/FOUR-6485.
- Set package-webentry to [bugfix/FOUR-6485](https://github.com/ProcessMaker/package-webentry/pull/150).
- Set package-collections to patch-1-cartegraph.
- Set package-savedsearch to patch-1-cartegraph.

## Issue(s) & Reproduction Steps

**Note**: This branch fixes two (2) tickets in Jira.

For [FOUR-6485](https://processmaker.atlassian.net/browse/FOUR-6485):
1. Create a custom webentry URL.
2. Copy the standard webentry URL, then switch it over to a custom URL and save (as well as save the process).
3. Visit the standard webentry URL.
4. Notice the standard URL still works as well as the custom webentry URL.

For [FOUR-6484](https://processmaker.atlassian.net/browse/FOUR-6484):

- Create a custom URL with a segment and one or more URL variables.
- Save the webentry URL and save the processes.
- Attempt to visit the URL without the expected variable(s). 
- You’ll notice you receive a 500 error, which is the HTTP response code for internal server error.

## Solution
- When a standard webentry URL is visited, there is now a check to see if a custom webentry url exists. If one does exists, then it returns a 404 error and exits (since the user should be using the custom url, not the standard).
- The AuthenticateSession middleware was extended from the framework to catch an uncaught exception which was thrown in specific situations using webentry, including when visiting a custom webentry url which expects variables that aren't present.

## How to Test
- When visiting a standard webentry url that has a custom url set, you should now encounter a 404.
- When visiting a custom webentry route that expects variables (but the url you're using doesn't have them), you should npw encounter a 404 instead of a 500.

## Related Tickets & Packages
- Ticket: https://processmaker.atlassian.net/browse/FOUR-6485
- Ticket: https://processmaker.atlassian.net/browse/FOUR-6484
- PR: https://github.com/ProcessMaker/package-webentry/pull/150

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Http/Kernel.php
+++ b/ProcessMaker/Http/Kernel.php
@@ -32,16 +32,13 @@ class Kernel extends HttpKernel
             \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
             \Illuminate\Session\Middleware\StartSession::class,
             \ProcessMaker\Http\Middleware\SessionStarted::class,
-            \Illuminate\Session\Middleware\AuthenticateSession::class,
+            \ProcessMaker\Http\Middleware\AuthenticateSession::class,
             \Illuminate\View\Middleware\ShareErrorsFromSession::class,
             //\ProcessMaker\Http\Middleware\VerifyCsrfToken::class,
             \ProcessMaker\Http\Middleware\SetLocale::class,       // This is disabled until all routes are handled by our new engine
             \Illuminate\Routing\Middleware\SubstituteBindings::class,
             \ProcessMaker\Http\Middleware\GenerateMenus::class,
             \Laravel\Passport\Http\Middleware\CreateFreshApiToken::class,
-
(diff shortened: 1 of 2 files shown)


agustinbusso (approved):
Working for both scenarios for ticket FOUR-6484 and FOUR-6485

**FOUR-6484**

https://user-images.githubusercontent.com/90727999/177864904-98cba262-c482-4bd3-96d7-09c0dd41a936.mov

**FOUR-6485**

https://user-images.githubusercontent.com/90727999/177864972-959d1cdd-deea-4004-890d-e8163cd9f1dc.mov

ryancooley (approved):

PR #4441, opened 2022-07-14
FOUR-6311: Added missing config value to allow java SDK compilation to run

runyan-co:
## Issue & Reproduction Steps
When installing processmaker/docker-executor-java, the java SDK build fails which then causes the entire docker image build to fail as well.

To replicate this issue:
1. Run `composer require processmaker/docker-executor-java`
2. Now run `php artisan docker-executor-java:install`

You'll notice compilation errors in the output and an exit code of 1.

## Solution
- Added an additional compiler configuration value for the java sdk compiler.

## How to Test
You should now be able to run the same steps as described above and build the docker-executor-java Docker image without errors and with an exit code of 0.

## Related Tickets & Packages
- Jira [FOUR-311 ticket](https://processmaker.atlassian.net/browse/FOUR-6311)

## Code Review Checklist

### DIFF
--- a/ProcessMaker/BuildSdk.php
+++ b/ProcessMaker/BuildSdk.php
@@ -1,13 +1,14 @@
 <?php
+
 namespace ProcessMaker;
 
-use \Exception;
-use \ZipArchive;
-use function GuzzleHttp\json_decode;
+use DOMDocument;
+use Exception;
 use ProcessMaker\Events\BuildScriptExecutor;
 use ProcessMaker\Facades\Docker;
 
-class BuildSdk {
+class BuildSdk
+{
     private $debug = true;
     private $image = "openapitools/openapi-generator-cli";
     private $tag = "v5.1.1";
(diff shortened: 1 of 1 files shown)


caleeli:
@runyan-co 
I could not replicate the issue reported. It seams to be a problem with the generated Swagger Classes for Java, but probably they are different now from the ones when the ticket was reported (4.2-x-develop).

The option you added to compiler `useIncrementalCompilation` (as described in https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html) could cause problems because it uses the datetime of the files to compile or not compile the changed java files. 

> 
> **useIncrementalCompilation**
> to enable/disable incremental compilation feature.
> 
> This leads to two different modes depending on the underlying compiler. The default javac compiler does the following:
> 
> true (default) in this mode the compiler plugin determines whether any JAR files the current module depends on have changed in the current build run; or any source file was added, removed or changed since the last compilation. If this is the case, the compiler plugin recompiles all sources.
> false (not recommended) this only compiles source files which are newer than their corresponding class files, namely which have changed since the last compilation. This does not recompile other classes which use the changed class, potentially leaving them with references to methods that no longer exist, leading to errors at runtime.
> 

Also, if you look at the minute `1:00` of the Java building video attached in the ticket, there is also a pusher error:
image

PR #4541, opened 2022-10-25
Hiding sidebar if empty - PM4EH-98

paulosakamoto:
## Issue & Reproduction Steps
It is necessary to hide the sidebar in some routes if the user is not an administrator.

It was working until version PM 4.1.27

This issue was reported in [PM4EH-98](https://processmaker.atlassian.net/browse/PM4EH-98).

## Solution
The only modification made in this project is if the `$sidebar` has the `display === false` option the sidebar will not be displayed.

It will be up to the other project ( [adps](https://github.com/ProcessMaker/package-zj-adps) ) to decide in which situations the menu items should be hidden.

## How to Test
- Install the package [adps](https://github.com/ProcessMaker/package-zj-adps), branch bugfix/PM4EH-98
- Create a non-admin user and give full access to requests, tasks and users
- Navigate between the Request, Task and Admin tabs.
- A non-admin should not have a sidebar on the Requests and Tasks tabs, but should see the sidebar on the Admin tab

## Related Tickets & Packages
- [PM4EH-98](https://processmaker.atlassian.net/browse/PM4EH-98)

## Code Review Checklist

### DIFF
--- a/resources/views/layouts/layout.blade.php
+++ b/resources/views/layouts/layout.blade.php
@@ -83,9 +83,11 @@
 <body>
 <a class="skip-navigation alert alert-info" role="link" href="#main" tabindex="1">{{ __('Skip to Content') }}</a>
 <div class="d-flex w-100 mw-100 h-100 mh-100" id="app-container">
+  @hasSection('sidebar')
   <div id="sidebar" class="d-print-none" :class="{expanded: expanded}">
       @yield('sidebar')
   </div>
+  @endif
   <div class="d-flex flex-grow-1 flex-column overflow-hidden">
     <div class="flex-grow-1">
         @include('layouts.navbar')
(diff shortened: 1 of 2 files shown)


danloa (commented):
@paulosakamoto  I've being testing the solution but I see that probably there is a more complex problem: Laravel processes package Middlewares in alphabetical order. 

I tested with data-sources, saved-search, and translations packages. 
If the ADPS middleware code is put in data-sources, saved-search overwrites those changes and the side bar menus are displayed.

When I put the ADPS middleware in translations, it works ok because translations overwrites saved-search changes and the menus are removed.

As the package is called ADPS, it is processed first by Laravel and the subsequent packages overwrite its changes...

In short: we need to investigate so that middlewares are processed in an expected order (May be to implement a "order position" variable in Middlawares?) Probably Laravel has a way to implement this.

paulosakamoto:
The implementations will be done only in the adps package.

PR #4546, opened 2022-10-26
Fix failing test related to MigrateFresh - FOUR-6933

paulosakamoto:
## Issue & Reproduction Steps
In [MigrateFresh:35](https://github.com/ProcessMaker/processmaker/blob/develop/ProcessMaker/Console/Commands/MigrateFresh.php#L35), 
the migrations table is being manually created to resolve the bug described in this [PR]( https://github.com/ProcessMaker/processmaker/pull/4537).

## Solution
Add a configuration so that the migrations table doesn't try to be created.

## How to Test
Run this test:
```
php vendor/phpunit/phpunit/phpunit --colors --stop-on-failure tests/Feature/Api/ChangePasswordTest.php
```

## Related Tickets & Packages
- [FOUR-6933](https://processmaker.atlassian.net/browse/FOUR-6933)

## Code Review Checklist

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -95,10 +95,7 @@ if (env('TEST_TOKEN')) {
     Artisan::call('db:wipe', ['--database' => \DB::connection()->getName()]);
     Artisan::call('migrate:fresh', []);
     Artisan::call('db:seed', ['--class' => 'AnonymousUserSeeder']);
-
-    // In \ProcessMaker\Console\Commands\MigrateFresh:35
-    // The migrations table is being manually created to resolve the bug described below.
-    // https://github.com/ProcessMaker/processmaker/pull/4537
+    
     \Illuminate\Foundation\Testing\RefreshDatabaseState::$migrated = true;
 
     ScriptExecutor::firstOrCreate(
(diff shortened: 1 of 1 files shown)


nolanpro (approved):

tests/bootstrap.php
```diff
@@ -96,6 +96,11 @@
     Artisan::call('migrate:fresh', []);
     Artisan::call('db:seed', ['--class' => 'AnonymousUserSeeder']);
 
+    // In \ProcessMaker\Console\Commands\MigrateFresh:35
+    // The migrations table is being manually created to resolve the bug described below.
+    // https://github.com/ProcessMaker/processmaker/pull/4537
```
ryancooley, line 101: @paulosakamoto I appreciate the comments but I think we should remove them so that we don't overwhelm our code with references to tickets. If we ever need to trace back this addition to the codebase we can use `git blame` to do so. Thank you!

ryancooley (changes requested):

tests/bootstrap.php
paulosakamoto: Done.

ryancooley (approved):

PR #4576, opened 2022-11-23
FOUR-7110: Improved request retry

runyan-co:
## Issue & Reproduction Steps
This PR covers the issue(s) covered in FOUR-7110, namely a 422 error which occurred when the user rapidly used the "Retry" button to trigger the retry request feature. 

## Solution
- Disables the retry button after initial click
- Switches request to "ACTIVE" upon retry
- Switched to better error logging/reporting for when script tasks fail

## How to Test
You should only be able to retry a request once until the page reloads. If the request is in a "In Progress" state, the retry button will not appear (as this is expected). If you refresh again and a script has errored out, it should update the request and you should be able to retry it again.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-7110
- https://processmaker.atlassian.net/browse/FOUR-6832

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Jobs/RunScriptTask.php
+++ b/ProcessMaker/Jobs/RunScriptTask.php
@@ -105,13 +105,7 @@ class RunScriptTask extends BpmnAction implements ShouldQueue
                 $this->instance = $instance;
             });
         } catch (Throwable $exception) {
-            // Change to error status
-            $token->setStatus(ScriptTaskInterface::TOKEN_STATE_FAILING);
-            $error = $element->getRepository()->createError();
-            $error->setName($exception->getMessage());
-            $token->setProperty('error', $error);
-            Log::error('Script failed: ' . $scriptRef . ' - ' . $exception->getMessage());
-            Log::error($exception->getTraceAsString());
+            $token->logError($exception, $element);
         }
     }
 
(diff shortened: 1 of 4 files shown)


ProcessMaker/Jobs/RunScriptTask.php
```diff
@@ -105,13 +105,7 @@ public function action(ProcessRequestToken $token = null, ScriptTaskInterface $e
...
             });
         } catch (Throwable $exception) {
-            // Change to error status
-            $token->setStatus(ScriptTaskInterface::TOKEN_STATE_FAILING);
-            $error = $element->getRepository()->createError();
-            $error->setName($exception->getMessage());
-            $token->setProperty('error', $error);
-            Log::error('Script failed: ' . $scriptRef . ' - ' . $exception->getMessage());
-            Log::error($exception->getTraceAsString());
+            $token->logError($exception, $element);
```
danloa, line 108: It seems that when the old lines were removed (108 to 114) when a script task within  a subprocess fails, it is not marked as failing and is left in status ACTIVE. So we have script/service task that never finish or fail and the Retry button  is not  displayed.
The next attached code can be used to reproduce what I describe (Use the start event "Start Event Sub")

[2_scripts.zip](https://github.com/ProcessMaker/processmaker/files/10078479/2_scripts.zip)

runyan-co: @danloa Good catch! The only solution I could come up with here was to re-add the lines I took out and I also added in line 115, which seems to work correctly.

Can you confirm? and if it isn't working correctly, what would you recommend? I ask because I'm a bit out of my depth here in working with the more BPMN-intensive portion of the app. :)

danloa (approved):

ryancooley (approved):

PR #4618, opened 2023-02-03
Set processmaker_intended cookie to sameSite=none

nolanpro:
## Issue & Reproduction Steps
See https://processmaker.atlassian.net/browse/FOUR-7350

## Solution
The processmaker_intended cookie needs to be set with the options sameSite=none and secure=true.

This is because the SSO login post originated from a 3rd party domain (In this case, the SAML IDP) and chrome doesn't allow that unless sameSite is none. 

image

It looks like in Laravel 7 they [changed some defaults](https://github.com/laravel/framework/commit/f81b6ed01fb60580ade8c7fb4386aff4cb4d7719#diff-8ac2d435acb71caedbdd807ad9fb43e4c22e62db958e41b7ba91029bfd76f192). 

I didn't see anything in the release notes about it.

## How to Test
See reproduction steps in the the issue. You need to set up SSO (I think any provider will have the same issue). You only need an authenticated web entry and you can directly go to the web entry URL in an incognito window. Login with SSO and you should be redirected to the authenticated web entry from.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-7350

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Http/Controllers/Auth/LoginController.php
+++ b/ProcessMaker/Http/Controllers/Auth/LoginController.php
@@ -59,7 +59,7 @@ class LoginController extends Controller
         // clear cookie to avoid an issue when logout SLO and then try to login with simple PM login form
         \Cookie::queue(\Cookie::forget(config('session.cookie')));
         // cookie required here because SSO redirect resets the session
-        $cookie = cookie('processmaker_intended', redirect()->intended()->getTargetUrl(), 10, '/');
+        $cookie = cookie('processmaker_intended', redirect()->intended()->getTargetUrl(), 10, null, null, true, true, false, 'none');
         $response = response(view('auth.login', compact('addons', 'block')));
         $response->withCookie($cookie);
 


ProcessMaker/Http/Controllers/Auth/LoginController.php
```diff
@@ -59,7 +59,7 @@ public function showLoginForm()
         // clear cookie to avoid an issue when logout SLO and then try to login with simple PM login form
         \Cookie::queue(\Cookie::forget(config('session.cookie')));
         // cookie required here because SSO redirect resets the session
-        $cookie = cookie('processmaker_intended', redirect()->intended()->getTargetUrl(), 10, '/');
+        $cookie = cookie('processmaker_intended', redirect()->intended()->getTargetUrl(), 10, null, null, true, true, false, 'none');
```
caleeli, line 62: @nolanpro is there a reason why you changed the cookie path from '/' to null?

nolanpro: @caleeli It's not needed. The default set by config/session.php is '/' so it's the same thing. It's better to use the default values.

caleeli: @nolanpro According to the https://www.php.net/manual/en/function.setcookie.php, the default path for the cookies is "_the current directory that the cookie is being set in_". But yes! in this case the directory of the `/login` page is `/` and that is the same.

On the other hand, 🤔 I understand that the `config/session.php` set the parameters for the cookie `processmaker_session`:

image

caleeli (approved):

PR #4695, opened 2023-04-20
[FOUR-8126] Recompiling Mix fails on instance when colors have been customized

ryancooley:
### Steps to Reproduce
1. Run `npm run prod` or `npm run dev` on a clean install with no custom CSS colors, ensuring that it runs as expected
2. Visit `/admin/customize-ui`
3. Change some of the default colors and click save, allowing the CSS to regenerate
4. Run `npm run prod` or `npm run dev` again

### Current Behavior

The build fails with the following error:
```
SassError: Undefined variable.
  ╷
6 │   color: $darkneutral;
  │          ^^^^^^^^^^^^
```

### Expected Behavior
The build should complete without error.

## Solution
- Removed the color variables that were added to support the new Import/Export functionality from the _colors.scss file and moved them to the _variables.scss file. Since the Customize UI functionality was replacing the _colors.scss file without those new colors in it, the build would fail. Now, those color variables exist regardless of the status of the _colors.scss file, so the build runs as expected.
- During testing it was also discovered that the default colors presented on the Customize UI page were based on the old default colors which have since changed. So those were updated as well.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-8126

## Code Review Checklist

### DIFF
--- a/resources/js/admin/cssOverride/components/SiteDesign.vue
+++ b/resources/js/admin/cssOverride/components/SiteDesign.vue
@@ -226,32 +226,32 @@ export default {
       colorDefault: [
         {
           id: '$primary',
-          value: '#3397E1',
+          value: '#0872C2',
           title: this.$t('Primary')
         },
         {
           id: '$secondary',
-          value: '#788793',
+          value: '#6C757D',
           title: this.$t('Secondary')
         },
         {
           id: '$success',
-          value: '#00BF9C',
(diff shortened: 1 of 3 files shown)


agustinbusso (approved):
I could compile after the changes. Colors of the new version are correctly applied after using "reset" functionality

boliviacoca (approved):

PR #4735, opened 2023-05-03
FOUR-8342 Duplicate media queries

caleeli:
## Issue & Reproduction Steps
208 duplicate queries; not huge savings but unnecessary DB load If don't need avatar on every API query, don't append it
User:129

## Solution
- Convert `avatar` into a real column with URL of the avatar to avoid the query that constantly gets the avatar for each user every time an User is loaded.
- This URL is updated when the User profile is updated

## How to Test
- Download the branch with the change.
- Run the migrations
- Run the artisan command `php artisan users:populate-avatar`
- Test the Process Maker User Avatars, changing avatars, see the proper avatar is displayed in each place:
  - List of Processes
  - List of requests
  - Filters
  - Open Request
  - Open Task
  - Participants in Saved Searches

## Important step for CloudOps
- Run the migrations to add the avatar column
- Run the artisan command `php artisan users:populate-avatar` to populate with the current avatar URLs

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-8342

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Console/Commands/PopulateAvatarColumn.php
+++ b/ProcessMaker/Console/Commands/PopulateAvatarColumn.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace ProcessMaker\Console\Commands;
+
+use Illuminate\Console\Command;
+use ProcessMaker\Models\User;
+
+class PopulateAvatarColumn extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'users:populate-avatar';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Populate the avatar column with the current media files';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        // Fetch all users
+        $users = User::cursor();
+
+        // Iterate through each user and populate the avatar column
+        foreach ($users as $user) {
+            $user->avatar = $user->getAvatar();
+            $user->save();
+        }
+
+        // Output a message indicating success
+        $this->info('Avatar column populated successfully.');
+
+        return 0;
+    }
+}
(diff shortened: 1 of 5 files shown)


runyan-co (changes requested):
Hey @caleeli, this looks great! The only change I'd recommend is adding a call to the new populate avatar command to an upgrade migration so it's called when upgrading to a new version. Would you be able to add that in?

caleeli:
> Hey @caleeli, this looks great! The only change I'd recommend is adding a call to the new populate avatar command to an upgrade migration so it's called when upgrading to a new version. Would you be able to add that in?

Thanks @runyan-co for the feedback, the suggestion was implemented

runyan-co (changes requested):
Hey @caleeli, thanks for the changes! I do believe we need to add them to an [_upgrade migration_ specifically](https://github.com/ProcessMaker/processmaker/blob/develop/upgrades/2022_04_19_164837_sanitize_usernames.php), which you can generate an empty one to use just like a regular migration (`php artisan make:upgrade`).  Do ignore the `$to` property as it was deprecated in the upgrade migrations. Let me know if I can answer any questions about this! 👍🏻

nolanpro:
@caleeli @runyan-co I also think its a good idea to only have migrations in the migrations folder. If we ever want to squash migrations to improve performance (`php artisan schema:dump --prune`), having artisan commands in there would make that difficult.

I realize we already do it in older migration files but I think it would be good practice going forward to only have DB migrations (that can be directly converted to SQL) in the migrations folder.

runyan-co:
> @caleeli @runyan-co I also think its a good idea to only have migrations in the migrations folder. If we ever want to squash migrations to improve performance (`php artisan schema:dump --prune`), having artisan commands in there would make that difficult.
> 
> I realize we already do it in older migration files but I think it would be good practice going forward to only have DB migrations (that can be directly converted to SQL) in the migrations folder.

I agree completely @nolanpro, the upgrade migration should only contain the call to the `users:populate-avatar` artisan command without any schema changes. This was a feature we specifically designed to do this type of thing since standard migrations are meant for db/schema related changes.

runyan-co (approved):

nolanpro (approved):

ryancooley (approved):

PR #5021, opened 2023-07-19
FOUR-9353 Fix queue connection timeout

caleeli:
## Issue & Reproduction Steps
When engine is configured with rabbitmq and it is not used for a long time the connection to the message broker times out.

## Solution
- Setup the keepalive and timeout properties for the connection to keep alive the connection.

## How to Test
1. Run a process and wait a couple of days (do not run or complete any action in the server)
2. then check the connection is still alive by running another processes.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-9353

## Code Review Checklist


julceslauhub (approved):
Approved

PR #5039, opened 2023-07-20
FOUR-9412 Fix retry script task

caleeli:
## Issue & Reproduction Steps
During the retry test, a failed script task that was not fixed, changed the process state to ACTIVE when the user pressed "RETRY" instead of remaining in ERROR as intended.

## Solution
- Change to ACTIVE only when the script task is completed successfully.

## How to Test
- Create a script task with a throw exception
- Add the script to a process
- Run the process
- The task fails, then try RETRY, the request should keep in ERROR status
- Fix the script
- try RETRY again
- The request should continue (to ACTIVE if there is a next task or COMPLETED if it is the last)

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-9412

## Code Review Checklist


julceslauhub (approved):
Approved

julceslauhub (approved):
Approved

PR #5062, opened 2023-07-31
Fix issue when pm-blocks-package is not installed

eiresendez:
## Issue & Reproduction Steps

When the PM Block package is not installed and we attempt to access the PM Block associated with a Process, a PHP Fatal error is thrown because the class does not exist.

## Solution
-  An additional check is added to verify that the PM Block package is installed.

## How to Test
- Remove PM Block Package
- Access the route `/modeler/{id}`.

## Related Tickets & Packages
- [FOUR-9471](https://processmaker.atlassian.net/browse/FOUR-9471)

[FOUR-9471]: https://processmaker.atlassian.net/browse/FOUR-9471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ProcessMaker/Policies/ProcessPolicy.php
+++ b/ProcessMaker/Policies/ProcessPolicy.php
@@ -3,7 +3,6 @@
 namespace ProcessMaker\Policies;
 
 use Illuminate\Auth\Access\HandlesAuthorization;
-use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


estebangallego (approved):
Approved.

PR #5107, opened 2023-08-07
Allow null run_as_user_id in scripts export

nolanpro:
## Issue & Reproduction Steps
See https://processmaker.atlassian.net/browse/FOUR-9646?focusedCommentId=339247

We now allow null values for run_as_user_id (from [9747](https://processmaker.atlassian.net/browse/FOUR-9747)) however a script with null run_as_user_id is not able to export.

## Solution
- Check for a null run_as_user_id before adding the user as a dependent in the script exporter.

## How to Test
- Create a script with a regular user as the Run Script As user
- Create process with the script
- Export the process
- Clear your DB or use another instance that does not have the user
- Import the process (this should work fine, as it did before this fix. The run_as_user_id will be null in the database)
- Create a PM block form the imported process. Before this fix we would get an error.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-9646

## Code Review Checklist

### DIFF
--- a/ProcessMaker/ImportExport/Exporters/ScriptExporter.php
+++ b/ProcessMaker/ImportExport/Exporters/ScriptExporter.php
@@ -20,7 +20,9 @@ class ScriptExporter extends ExporterBase
             $this->addDependent(DependentType::ENVIRONMENT_VARIABLES, $environmentVariable, EnvironmentVariableExporter::class);
         }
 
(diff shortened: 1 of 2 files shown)


agustinbusso (approved):
The fix is correct:

https://github.com/ProcessMaker/processmaker/assets/90727999/89055bf9-f184-4638-a8ad-43767b9f4e9b

PR #5156, opened 2023-08-15
FOUR-9964 | Template Search Behaving Erratically

mcraeteisha:
# Issue
Ticket: [FOUR-9964](https://processmaker.atlassian.net/browse/FOUR-9964)

When searching for a template in the New Process Modal, the search functionality is behaving erratically.

# Solution
- There was a watcher on the search filter that fetched templates with each change of the filter.
	- Search functionality works without the watcher. However, the watcher is needed to fetch templates when the filter is an empty string (i.e. when a user searches and then backspaces/clears our their search term). Added a conditional to the watcher for this case.
- This PR also adds functionality to the search button in the search bar when clicked.
- This PR also removes the extra spacing above the 'No Data Available' search result so that users don't have to scroll to see the image when there are no results.

# How to Test
1. Go to branch `bugfix/FOUR-9964` in `processmaker`.
2. Go to Designer → Processes → +PROCESS.
3. Search for templates.
	- The functionality should not behave erratically.
    - If searching for a template that doesn't exist, you should not have to scroll to see the 'No Data Available' result.

# Code Review Checklist

- [ ]  This code adheres to [ProcessMaker Coding Guidelines](https://github.com/ProcessMaker/processmaker/wiki/Coding-Guidelines).

[FOUR-9964]: https://processmaker.atlassian.net/browse/FOUR-9964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/resources/js/components/templates/TemplateSearch.vue
+++ b/resources/js/components/templates/TemplateSearch.vue
@@ -3,7 +3,7 @@
     <div class="pb-3">
       <b-input-group v-if="component === 'template-select-card'">
         <b-input-group-prepend>
-          <b-btn class="btn-search-run px-2" :title="$t('Search Templates')">
+          <b-btn class="btn-search-run px-2" :title="$t('Search Templates')" @click="fetch()">
             <i class="fas fa-search search-icon" />
           </b-btn>
         </b-input-group-prepend>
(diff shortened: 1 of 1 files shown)


eiresendez (changes requested):
hello @mcraeteisha there are still some issues where if a word is misspelled, it stops searching:

https://github.com/ProcessMaker/processmaker/assets/90741874/89ec2ee5-808a-4f7b-9621-5842990a1712

Also, if it's not too complicated and if you find it useful... could you add a debounce to the fetch function to improve the UX? Since an API call is made every time the user types a letter, the response might not be quick and could result in search results that lag several characters behind what has been typed.. with debounce, the search would be executed once the user has finished typing, example:

```
created() {
  // Create a "debounced" version of the fetch() function, maybe using lodash.debounce()
  this.fetchDebounced = debounce(this.fetch, 300);
},
watch: {
  filter() {
    this.fetchDebounced();
  },
},
```

eiresendez (approved):
@mcraeteisha the code was making 2 API calls. I removed the `filter()` from the `watch` since it was already defined in `datatable.js` (with a debounced version). Additionally, I relocated the `loading` flag. Please review it tomorrow, thanks!

PR #5162, opened 2023-08-15
Rich text-html settings stopped working to initiate a requests

marcoAntonioNina:
## Issue & Reproduction Steps
the problem is replicated when using NayraDocker when you want to start a request by URL
http://192.168.1.205/process_events/2?event=node_21

the _request variable is not defined.

## Solution
- as the request id is needed we no longer use the _request variable

## How to Test
try to start a case by url
http://192.168.1.205/process_events/2?event=node_21

## Related Tickets & Packages
[FOUR-9840](https://processmaker.atlassian.net/browse/FOUR-9840)

## Code Review Checklist

[FOUR-9840]: https://processmaker.atlassian.net/browse/FOUR-9840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ProcessMaker/Http/Controllers/ProcessController.php
+++ b/ProcessMaker/Http/Controllers/ProcessController.php
@@ -246,11 +246,10 @@ class ProcessController extends Controller
 
     public function triggerStartEventApi(Process $process, Request $request)
     {
-        $api_request = new ApiProcessController();
-        $response = $api_request->triggerStartEvent($process, $request);
-        $instance_id = $response->data['_request']['id'];
+        $apiRequest = new ApiProcessController();
+        $response = $apiRequest->triggerStartEvent($process, $request);
 
-        return redirect('/requests/' . $instance_id);
+        return redirect('/requests/' . $response->id);
     }
 
     private function checkAuth()


ProcessMaker/Http/Controllers/ProcessController.php
```diff
@@ -246,11 +246,10 @@ public function destroy(Process $process) // destory existing process to DB / UI
...
     public function triggerStartEventApi(Process $process, Request $request)
     {
-        $api_request = new ApiProcessController();
-        $response = $api_request->triggerStartEvent($process, $request);
-        $instance_id = $response->data['_request']['id'];
+        $apiRequest = new ApiProcessController();
+        $response = $apiRequest->triggerStartEvent($process, $request);
 
-        return redirect('/requests/' . $instance_id);
+        return redirect('/requests/' . $response->id);
```
caleeli, line 252: I agree with this change.

About the magic variable `data['_request']` since Nayra-engine runs always in async mode (including the Start Event), this magic variable will not be populated immediatly. Right now, the in-core Engine runs the Start Event using a `distpachNow` but this was deprecated in Laravel 8.

Considering that scenario in this case is right to use `$request->id` since this is the only one value required from `_request`.

PR #5164, opened 2023-08-16
Feature/FOUR-9922 Json Crack Viewer

gustavobascope:
## Issue & Reproduction Steps
Integrate the library [JSON Crack - Visualize Data to Graphs](https://jsoncrack.com/docs) into the overlay window

## Solution
Added a Json Crack widget on a Iframe

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-9922


nolanpro (changes requested):
@gustavobascope This looks really good! A few small changes:
- Can we change the path from `/jsoncrack/` to `/json-browser/` in case we decide to use a different library later? I think there is a build option in jsoncrack for that.
- I don't think we need the change in web.php. When I use `src="/jsoncrack/widget.html"` instead of `src="/jsoncrack/widget"` for the iframe it works for me

tokensPM:
QA server was successfully deployed https://ci-ec1ed2e38f.eng.processmaker.net

nolanpro (approved):

PR #5339, opened 2023-09-19
FOUR-10634 Imports fix

josechirivella:
Added index files to the components folder and Mixins to be exported as external packages to be used or consumed by our packages.

The SharedComponents index didn't change, but I changed the target in the `bootstrap.js` to target the upper-level index file (newly created) to export all these components.
I added a note to a component with the same content as in the Shared folder but many files are using it and I didn't want to introduce so many changes, but be free to change those if you have the time.

What I would do is the following. Go to the savedsearch package, look at the broken imports, check if we already have a index file or if those components/mixins are being exported, and change the name of the import. If those components/mixins aren't being exported, create an `index.js` and start exporting those. Remember to add the import in the `bootstrap.js` located in the root `resources` folder. 

Added a change into the laravel mix to externalize all the dependencies coming from the package.json @ryancooley. Let me know if I should take out this change and create a new PR for it. 

### Depends on
* https://github.com/ProcessMaker/package-savedsearch/pull/381

ci:next
.


josechirivella:
Closing since this was resolved in https://github.com/ProcessMaker/package-savedsearch/pull/403

PR #5499, opened 2023-10-18
feature/FOUR-1200: Add a new Filter overdue in the API  "/tasks"

pmPaulis:
## Issue & Reproduction Steps
Add a new Filter overdue in the API  "/tasks"

## Solution
- We need to implement a new filter over the API `/tasks` in order to filter the overdue
- 
image

## How to Test
a new parameter was added overdue=true in order to show only the overdue request
`api/1.0/tasks?page=1&include=process,processRequest,processRequest.user,user,data&pmql=(user_id = 1) AND (status = "In Progress")&per_page=10&order_by=ID&order_direction=DESC&overdue=true`

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-11200

## Code Review Checklist

ci:next

### DIFF
--- a/ProcessMaker/Http/Controllers/Api/TaskController.php
+++ b/ProcessMaker/Http/Controllers/Api/TaskController.php
@@ -213,6 +213,9 @@ class TaskController extends Controller
             return $query->count();
         }
 
+        // Apply filter overdue
+        $query->overdue($request->input('overdue'));
+
         try {
             $response = $this->handleOrderByRequestName($request, $query->get());
         } catch (QueryException $e) {
--- a/ProcessMaker/Http/Controllers/Api/TaskController.php
+++ b/ProcessMaker/Http/Controllers/Api/TaskController.php
@@ -102,9 +102,7 @@ class TaskController extends Controller
         }
 
         $query = ProcessRequestToken::with(['processRequest', 'user']);
-
         $query->select('process_request_tokens.*');
-
         $include = $request->input('include') ? explode(',', $request->input('include')) : [];
 
         if (in_array('data', $include)) {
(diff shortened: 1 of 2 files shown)


ProcessMaker/Http/Controllers/Api/TaskController.php
```diff
@@ -191,9 +202,9 @@ public function index(Request $request, $getTotal = false, User $user = null)
             try {
                 $query->pmql($pmql, null, $user);
             } catch (QueryException $e) {
-                return response(['message' => __('Your PMQL search could not be completed.')], 400);
+                abort('Your PMQL search could not be completed.', 400);
```
fagubla, line 205: Does this change affect the display of messages?

ProcessMaker/Http/Controllers/Api/TaskController.php
```diff
@@ -191,9 +202,9 @@ public function index(Request $request, $getTotal = false, User $user = null)
...
             } catch (SyntaxError $e) {
-                return response(['message' => __('Your PMQL contains invalid syntax.')], 400);
+                abort('Your PMQL contains invalid syntax.', 400);
```
fagubla, line 207: Does this change affect the display of messages?

fagubla (changes requested):

ProcessMaker/Http/Controllers/Api/TaskController.php
pmPaulis: This is a SonnarQueue observations because this funtion 5 return and only 3 are permitted

ProcessMaker/Http/Controllers/Api/TaskController.php
pmPaulis: This is a SonnarQueue observations because this funtion 5 return and only 3 are permitted

PR #5605, opened 2023-11-08
FOUR-11226 The column with the LDAP options is shown to the far right…

gproly:
…, crowding the buttons

**Steps to Reproduce:**

1. Log in 
2. Go To Admin 
3. Click on Settings
4. Click on LDAP  

**Current Behavior:** 

When trying to perform the LDAP settings, we see that the settings column is very far to the right, this causes the buttons for the copy, edit and delete functions to look crowded. watch video

**Expected Behavior:**

The configuration and options column should go a little further to the left

### DIFF
--- a/resources/js/admin/settings/components/SettingsListing.vue
+++ b/resources/js/admin/settings/components/SettingsListing.vue
@@ -7,28 +7,28 @@
       :ai-enabled="false"
       :aria-label="$t('Advanced Search (PMQL)')"
       @submit="onNLQConversion">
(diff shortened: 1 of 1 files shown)


marcoAntonioNina:
@gproly, I can see that the other tabs have lost your style.
image

marcoAntonioNina (changes requested):
please review all tabs
image

marcoAntonioNina (approved):

PR #5648, opened 2023-11-16
Docker timeouts can also have exit code 9

nolanpro:
## Issue & Reproduction Steps
On k8s, script timeouts are showing an exception trace instead of the timeout error that we had on ECS instances.

## Solution
- For some reason, on alpine linux, the timeout command can have the exit code `9` when run with php's `exec` command. This is different then on Mac and Ubutnu where it returns `137`. The solutions is to treat both exit codes as timeouts.

## How to Test
- On a k8s instance, set the timeout to 5 and run a script with `sleep(10);` to trigger the timeout out. It should show the timeout error instead and not the exception trace.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-10600

## Code Review Checklist


julceslauhub (approved):
Approved

PR #5700, opened 2023-11-24
FOUR-12567 - Problem with the in-flight map of a request

eiresendez:
## Issue & Reproduction Steps
Due to the fact that the status of the sequence flow is not recorded in the `process_request_tokens` table, difficulties arise in determining whether a flow sequence has been completed or not.

- To replicate this problem, import the attached process.

## Solution
- It is suggested to evaluate the condition of the sequence flow using the process data to determine if a flow sequence has been traversed.
  - Possibly, with more time and approval, another viable solution would be to record the states of the sequence flow in the database, similar to how it is handled with other nodes. This way, it would be much easier to identify if that route has been traversed.

## How to Test
- Follow the reproduction steps of above.

## Related Tickets & Packages
- [FOUR-12567](https://processmaker.atlassian.net/browse/FOUR-12567)

## Code Review Checklist

[FOUR-12567]: https://processmaker.atlassian.net/browse/FOUR-12567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ProcessMaker/Http/Controllers/Process/ModelerController.php
+++ b/ProcessMaker/Http/Controllers/Process/ModelerController.php
@@ -67,7 +67,7 @@ class ModelerController extends Controller
             'manager' => $manager,
             'signalPermissions' => SignalManager::permissions($request->user()),
             'autoSaveDelay' => config('versions.delay.process', 5000),
-            'isVersionsInstalled' => PackageHelper::isPackageInstalled('ProcessMaker\Package\Versions\PluginServiceProvider'),
+            'isVersionsInstalled' => PackageHelper::isPmPackageVersionsInstalled(),
             'isDraft' => $draft !== null,
             'pmBlockList' => $pmBlockList,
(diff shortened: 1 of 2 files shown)


gustavobascope (commented):
The solution presented a new scenario when the condition gateway returns to a task  than one time the previous sequence flow stops been highlithed.
Therefore, I suggest enhancing this by using the second approach, which involves storing a record of the sequence flow in the process_request_tokens table, prioritizing that task is essential, as it may require more time

image

eiresendez:
Closing this PR. It's recommended to record the state of a sequence flow in the `process_request_tokens` table, similar to how it is handled with the other nodes. This way, it will be much easier to identify if that path has been traversed and how many times it has been traveled.

PR #5708, opened 2023-11-28
Improve get activated transition in gateway

caleeli:
Fix way to send the reference to Gateway Transition was triggered or passed. This is used for comments.

- This code use the transition index to reference the outgoing chosen by a gateway instead of the ID.

## Code Review Checklist


ProcessMaker/Nayra/Repositories/PersistenceTokenTrait.php
```diff
@@ -165,9 +166,17 @@ public function persistGatewayTokenConsumed(array $transaction)
     public function persistGatewayTokenPassed(array $transaction)
     {
         $gateway = $this->deserializer->unserializeEntity($transaction['gateway']);
-        $transition = $this->deserializer->unserializeEntity($transaction['transition']);
+        if (!is_numeric($transaction['transition'])) {
+            Log::info('Invalid transition id for gateway token passed. ' . json_encode($transaction));
+            return;
+        }
+        $transition = $gateway->getTransitions()[$transaction['transition']] ?? null;
+        if (empty($transaction)) {
```
nolanpro, line 174: @caleeli should this be `empty($transition)` instead of `empty($transaction)`?

nolanpro (approved):

PR #5837, opened 2023-12-18
FOUR-12303 Migration/Upgrade to populate case numbers

caleeli:
## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-12303

## Code Review Checklist


upgrades/2023_12_14_160541_populate_process_requests_case_number.php
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Reverse the upgrade migration.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
```
julceslauhub, line 108: Question: This upgrade doesn't have rollback action?

upgrades/2023_12_14_160541_populate_process_requests_case_number.php
```diff
@@ -0,0 +1,110 @@
...
+
+            // Display the processing rate
+            $this->displayRate($caseNumber, $startTime, $count);
+        });
+
+        // Update the auto increment value of table case_numbers to the next case number
+        // ALTER TABLE case_numbers AUTO_INCREMENT = {$caseNumber};
+        DB::statement("ALTER TABLE case_numbers AUTO_INCREMENT = {$caseNumber};");
+
+        echo "\n";
```
julceslauhub, line 72: `echo PHP_EOL;`

upgrades/2023_12_14_160541_populate_process_requests_case_number.php
caleeli: That's right, it does not have a rollback action.

upgrades/2023_12_14_160541_populate_process_requests_case_number.php
caleeli: Fixed.

upgrades/2023_12_14_160541_populate_process_requests_case_number.php
caleeli: The current upgrade process is designed without a rollback feature, as it's not necessary under typical circumstances. In the event that the upgrade must be re-run, a traditional rollback isn't suitable because it wouldn't revert the 'case_numbers' to their previous state. Instead, if the upgrade is initiated again, it starts by clearing the 'case_numbers' table. This approach ensures that the case numbers are always generated in a sequential order, maintaining data integrity and consistency.

julceslauhub (approved):
Approved

devmiguelangel (approved):

PR #5905, opened 2024-01-03
Fix translations for screen builder

nolanpro:
See https://github.com/ProcessMaker/screen-builder/pull/1499


pmoskrojas (approved):
The content corresponds to the text for screen build multi language

PR #5911, opened 2024-01-03
FOUR-12985 Login failed empty, blocks the user on the second failed attempt

julceslauhub:
https://processmaker.atlassian.net/browse/FOUR-12985


resources/js/admin/settings/components/SettingText.vue
```diff
@@ -103,6 +103,10 @@ export default {
       this.$refs.input.focus();
     },
     onSave() {
+      if (this.setting.ui?.isNotEmpty && this.transformed == '') {
+        ProcessMaker.alert(this.$t("This value is invalid, please use another value."), "danger");
```
caleeli, line 107: In order to keep the UX style with other validations, consider to add a `<b-form-group>` and use the `:invalid-feedback` and `:state` instead of `ProcessMaker.alert`
https://bootstrap-vue.org/docs/components/form-group
For the message consider:
```
The current value is empty but a value is required. Please provide a valid value.
```

resources/js/admin/settings/components/SettingText.vue
```diff
@@ -103,6 +103,10 @@ export default {
       this.$refs.input.focus();
     },
     onSave() {
+      if (this.setting.ui?.isNotEmpty && this.transformed == '') {
```
caleeli, line 106: Consider to use strict equality
```suggestion
      if (this.setting.ui?.isNotEmpty && (this.transformed === '' || this.transformed === null)) {
```

caleeli (approved):

PR #5965, opened 2024-01-09
FOUR-13160 more than one open session is not allowed on the same device when device restriction is enable

devmiguelangel:
## Issue & Reproduction Steps
The last session is kept, but the previous one is killed

## Solution
- Allow multiple sessions on the same device with device restriction enabled

## How to Test
1. Log in
2. Create a new user
3. From admin user configure Device Restriction
    a. Go to the Admin tab
    b. Go to the settings option in the sidebar menu
    c. Go to the Session Control tab
    d. Go to Device Restriction and edit
    e. Enable Kill existing session
    f. Click on the save button
4. Log in with a new user in a device in a browser (for example chrome)
5. Log in with the same user in other browser on the same device (for example firefox)

## Related Tickets & Packages
[FOUR-13160](https://processmaker.atlassian.net/browse/FOUR-13160)

## Code Review Checklist

ci:next
ci:deploy

[FOUR-13160]: https://processmaker.atlassian.net/browse/FOUR-13160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ProcessMaker/Listeners/UserSession.php
```diff
@@ -58,12 +46,12 @@ public function handle(object $event): void
             $user->sessions()
                 ->where([
                     ['is_active', true],
+                    ['ip_address', '!=', $ip],
                 ])
-                ->where(function (Builder $query) use ($agentDevice, $agentDeviceType, $agentPlatform, $ip) {
+                ->where(function (Builder $query) use ($agentDevice, $agentDeviceType, $agentPlatform) {
```
caleeli, line 51: Seems this could fail with two similar devices but with different IP addresses.

image

Note that in this case the last login (id=13 for user=3 ip=192.168.1.163) is not killing the previous session (id=12 user=3 ip=192.168.1.119)

For the following configuration:

image

caleeli: Maybe you could consider to add a:
```
                        ->orWhere('ip_address', '!=', $ip)
```

devmiguelangel: It was restored

caleeli (approved):

PR #6108, opened 2024-01-29
observation/FOUR-13907 The status of a task self service is shown as in progress

gustavobascope:
## Issue & Reproduction Steps
The status of a task self service is shown as in progress

## Solution
Fix the condition for the Self Service status badge badge

## How to Test

Log in 
Have a process that the second task goes to Self Service
Start a new case
Go to tab task
Click on self service option

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-13907

ci:deploy
ci:next

### DIFF
--- a/resources/js/tasks/components/TasksList.vue
+++ b/resources/js/tasks/components/TasksList.vue
@@ -376,14 +376,15 @@ export default {
     formatStatus(props) {
       let color;
       let label;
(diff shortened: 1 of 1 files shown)


pmoskrojas (approved):
The modification solves the problem of filtering tasks with Self Service and In Progress status.

PR #6152, opened 2024-02-05
Refactor some parts of the column filtering functionality

nolanpro:
- Add badges to show the filters applied to the tasks and requests page
- Adds the filter configs to the column definitions so we can get rid of all the Alias overrides
- Applies aliases only when fetching results
- Makes process search a free-form text field instead of a list of dropdowns [FOUR-13977](https://processmaker.atlassian.net/browse/FOUR-13977)

Requires https://github.com/ProcessMaker/package-savedsearch/pull/408
ci:next

[FOUR-13977]: https://processmaker.atlassian.net/browse/FOUR-13977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gproly (approved):
Approved

PR #6157, opened 2024-02-05
FOUR-13275 When opening a request/task by clicking on it, then clicki…

gproly:


nolanpro (changes requested):
Can you move this to a middleware called `ProcessMaker/Http/Middleware/NoCache.php` and add it to the routes in routes/web.php
```
->middleware(NoCache::class);
```
I think this is a good temporary solution until we revisit this next release

nolanpro (approved):

PR #6227, opened 2024-02-15
QA Observation Bugfix/FOUR-14155: You cannot filter using "between" from the second page

CarliPinell:
## Solution
This ticket fixes issue with filter between in second page in Requests

## How to Test
Go to Requests, add between filter in STARTED column

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-14155

## Code Review Checklist
ci:deploy
ci:next


tokensPM:
QA server K8S was successfully deployed https://ci-a3330cff72.engk8s.processmaker.net

gproly:
Analyzing this change, this behavior should also be appearing in the 'Tasks' listing. Please review the 'page' property in files 'resources/js/requests/components/RequestsListing.vue' and 'resources/js/tasks/components/ListMixin.js'. The correct logic would be: 
- if the filter settings for any column change, then the 'page' property should be set to 1. If the filter settings remain the same, then 'page' should not change, enabling navigation between pages. 
The function you should analyze is 'resources/js/common/PMColumnFilterPopoverCommonMixin.js -> storeFilterConfiguration()'; this is where the configuration is saved. It is only necessary to consider the 'filters' variable in this function.

Please let's consider analyzing alternative solutions that Nolan can provide us with, thank you.

resources/js/requests/components/RequestsListing.vue
```diff
@@ -426,6 +427,13 @@ export default {
             },
           )
           .then((response) => {
+            if (response.data.data.length === 0 && this.fetchFlag === 0){
+              this.page = 1;
+              this.fetch();
+              this.fetchFlag = 1;
+            } else {
+              this.fetchFlag = 0;
+            }
```
gproly, line 436: This solution is incomplete. Please, it is necessary to review the logic of the task list. It is most likely that the same behavior is occurring when you apply a filter, it is searching on the current page being viewed. It is necessary to check the different pages.

gproly (changes requested):

resources/js/requests/components/RequestsListing.vue
nolanpro: @CarliPinell @gproly I agree. Please reset the page back to 1 when ever a filter changes

CarliPinell: Thank you very much Roly for the comments. However in 'Tasks' there is no problem with the filter on the column that contains date as Due Date.

I will review the files you tell me, for now this solution works well and solves the problem reported by QA.

I look forward to any suggestions or improvements that Nolan can give us. Unfortunately these errors are reported very late when there is not much time left to do a more in-depth analysis.

CarliPinell:
The problem that was detected is that when a filter is used on page 2 or subsequent pages, the result is sent to page 1, but the screen remains on page 2, showing that no results were found. But if you go back to page 1, the filter results are found.

tokensPM:
QA server K8S was successfully deployed https://ci-a3330cff72.engk8s.processmaker.net

nolanpro (dismissed):
@CarliPinell  I think you need to add previousAdvancedFilter to resources/js/tasks/components/TasksList.vue in the data() method

gproly (approved):

gproly:
I have added the variable 'previousAdvancedFilter' in file 'resources/js/tasks/components/TasksList.vue', so this would be approved. Thank you.

tokensPM:
QA server K8S was successfully deployed https://ci-a3330cff72.engk8s.processmaker.net

resources/js/requests/components/RequestsListing.vue
CarliPinell: Solved. Logic was changed.

PR #6339, opened 2024-03-05
Bugfix/FOUR-14325: Self Service Not Working in Tasks Menu

agustinbusso:
This PR solves https://processmaker.atlassian.net/browse/FOUR-14325

ci:next


nolanpro (approved):

ryancooley:
@agustinbusso What was the purpose of the fix format commit ca02c76? It looks like your original commit only modifies three files but then that commit modifies hundreds. Can we limit the scope of the PR?

agustinbusso:
> @agustinbusso What was the purpose of the fix format commit [ca02c76](https://github.com/ProcessMaker/processmaker/commit/ca02c76c20e4a3ad9da386e97e4f6c7587a8fd90)? It looks like your original commit only modifies three files but then that commit modifies hundreds. Can we limit the scope of the PR?

When I updated with next branch my php-cs-fixer asked me to fix the files with formatting errors. I can revert the fix format commit

PR #6382, opened 2024-03-11
FOUR-14395: Database Error: Table not found: 'project_assets' doesn't exist

eiresendez:
## Issue & Reproduction Steps
This error occurs when the `package-projects` is not installed and a non-administrator user account is used.

## Solution
Currently, there is a check to determine if the projects table exists in the database schema. The proposed solution is:
- If the projects table does not exist, then the Gate will return false by default.

The other changes are related to SonarQube. 
- The function inside the Gate had more than 4 returns
- The before() function had the error: "Vote methods should return at least once a negative response".

## How to Test
Since deploying a CI Server without the projects package is not possible, the best option for a functional test is to do it manually.

## Related Tickets & Packages
- [FOUR-14395](https://processmaker.atlassian.net/browse/FOUR-14395)

## Code Review Checklist

[FOUR-14395]: https://processmaker.atlassian.net/browse/FOUR-14395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ProcessMaker/Providers/AuthServiceProvider.php
+++ b/ProcessMaker/Providers/AuthServiceProvider.php
@@ -63,6 +63,9 @@ class AuthServiceProvider extends ServiceProvider
             if ($user->is_administrator) {
                 return true;
             }
+
+            // Let other policies handle the request.
+            return null;
         });
 
         try {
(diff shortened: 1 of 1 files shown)


estebangallego (approved):
Approved! 🍾

PR #6619, opened 2024-04-08
[SPRING] observation/FOUR-14686 Save Draft does not work when a Task was opened

gustavobascope:
## Issue & Reproduction Steps
Save Draft does not works when a Task was opened

## Solution
Added new autosave validation for new tasks

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-14686

ci:deploy
ci:next

### DIFF
--- a/resources/views/tasks/edit.blade.php
+++ b/resources/views/tasks/edit.blade.php
@@ -375,7 +375,7 @@
     const userHasAccessToTask = {{ Auth::user()->can('update', $task) ? "true": "false" }};
     const userIsAdmin = {{ Auth::user()->is_administrator ? "true": "false" }};
     const userIsProcessManager = {{ Auth::user()->id === $task->process?->manager_id ? "true": "false" }};
-    const screenFields = @json($screenFields);
+    var screenFields = @json($screenFields);
 
   </script>
     @foreach($manager->getScripts() as $script)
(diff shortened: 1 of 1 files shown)


tokensPM:
QA server K8S was successfully deployed https://ci-a58157bfc8.engk8s.processmaker.net

nolanpro (changes requested):
@gustavobascope I dont think this is going to work. If a script task runs before this, all the values of the script task will be included in the draft. `const screenFields = @json($screenFields);` comes from from the initial load, same as `const task = @json($task);` but since it's from a start event, there are no screen fields yet. Once the interstitial is done, we have a new Task but we are not updating the screenFields variable.

I think we should change screenFields to a `var` instead of a const, and updating it when the task changes.

gustavobascope:
> @gustavobascope I dont think this is going to work. If a script task runs before this, all the values of the script task will be included in the draft. `const screenFields = @json($screenFields);` comes from from the initial load, same as `const task = @json($task);` but since it's from a start event, there are no screen fields yet. Once the interstitial is done, we have a new Task but we are not updating the screenFields variable.
> 
> I think we should change screenFields to a `var` instead of a const, and updating it when the task changes.

Fixed observations

tokensPM:
QA server K8S was successfully deployed https://ci-a58157bfc8.engk8s.processmaker.net

nolanpro (approved):

PR #6680, opened 2024-04-12
Filter tasks by user if they are not an admin

nolanpro:
See https://processmaker.atlassian.net/browse/FOUR-14709

- Filter tasks for non-admin users and users that don't have the view-all-requests permission
- Remove unused API call that was causing an error message for non-admin users

ci:next


gproly (approved):
Approve

PR #6917, opened 2024-06-06
Bugfix/FOUR-14265: Drafts being lost after renaming the process

agustinbusso:
## Issue & Reproduction Steps

- Create a new process.
- Add some elements to the process (such as a start event, a couple of tasks, and an end event).
- You can refresh your screen to validate that your changes are still there.
- Go to the process configuration and change the name of the process.

**Current Behavior:**
All the changes will be lost

**Expected Behavior:**
The draft with the recent changes should still be there after renaming.

**Notes for QA:**
The information lost only occurs on unpublished versions. However, these changes exist in the version history as drafts before renaming the process.

## Solution
- When saving in configuration, now we get the last version of the process and we update the BPMN.
- The button in configuration changed from **Save** to **Save and publish**
- If a draft exist for the process, when clicking in Save and publish, we show a confirmation modal to let the user decide if must continue.

## How to Test
- Create a process.
- Go to modeler and do some changes (add elements), wait the autosave
- Go to the process configuration and change the process name and save

- Do the same test as before with draft Processes and Published processes
- Do the same test as before with one alternative as a draft
- Do the same test as before with two alternatives as a draft (modify only the B)
- Do the same test as before with two alternatives as a draft (modify only the A)
- Do the same test as before with two alternatives as a draft (modify both)

Ensure the processes and the alternatives keeps the changes after renaming the process.
Ensure the confirmation modal shows in the configuration when saving if the process is draft

## Related Tickets & Packages
- [FOUR-14265](https://processmaker.atlassian.net/browse/FOUR-14265)

## Code Review Checklist

ci:next

[FOUR-14265]: https://processmaker.atlassian.net/browse/FOUR-14265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/resources/views/processes/edit.blade.php
+++ b/resources/views/processes/edit.blade.php
@@ -195,7 +195,7 @@
                             </div>
                             <div class="d-flex justify-content-end mt-2">
                                 {!! Form::button(__('Cancel'), ['class'=>'btn btn-outline-secondary', '@click' => 'onClose']) !!}
-                                {!! Form::button(__('Save'), ['class'=>'btn btn-secondary ml-2', '@click' => 'onUpdate']) !!}
+                                {!! Form::button(__('Save and publish'), ['class'=>'btn btn-secondary ml-2', '@click' => 'onUpdate']) !!}
                             </div>
                         </div>
 
@@ -422,7 +422,7 @@
                             </div>
                             <div class="d-flex justify-content-end mt-2">
                                 {!! Form::button(__('Cancel'), ['class'=>'btn btn-outline-secondary', '@click' => 'onClose']) !!}
-                                {!! Form::button(__('Save'), ['class'=>'btn btn-secondary ml-2', '@click' => 'onUpdate']) !!}
+                                {!! Form::button(__('Save and publish'), ['class'=>'btn btn-secondary ml-2', '@click' => 'onUpdate']) !!}
                             </div>
                         </div>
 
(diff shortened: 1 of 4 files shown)


resources/views/processes/edit.blade.php
```diff
@@ -578,13 +585,29 @@ class="custom-control-input">
...
+                },
+                handleUpdate() {
                     this.resetErrors();
                     let that = this;
                     this.formData.cancel_request = this.formatAssigneePermissions(this.canCancel);
                     this.formData.edit_data = this.formatAssigneePermissions(this.canEditData);
                     this.formData.cancel_screen_id = this.formatValueScreen(this.screenCancel);
                     this.formData.request_detail_screen_id = this.formatValueScreen(this.screenRequestDetail);
                     this.formData.manager_id = this.formatValueScreen(this.manager);
+                    console.log(this.formData);
```
fagubla, line 610: Remove console

fagubla (changes requested):

resources/views/processes/edit.blade.php
agustinbusso: @fagubla Removed thanks!

fagubla (approved):

PR #6918, opened 2024-06-06
FOUR-15859: On Process End - Start a request for another selected process

rodriquelca:
## Issue & Reproduction Steps

Expected behavior: 

ent event must redirect to another process if have this configuration
Actual behavior: 

## Solution
- add another process redirect feature for end event

https://github.com/ProcessMaker/processmaker/assets/1401911/165fccb7-6aee-494c-a372-ffbab294c465

## How to Test
- create the process A
- create the process B
- assign to both process the same user to the first task
- in process A create a end event 
- add end event redirection as process
- select the process B
- select the start event B
- run the process

## Related Tickets & Packages
-  https://processmaker.atlassian.net/browse/FOUR-15859
- https://github.com/ProcessMaker/screen-builder/pull/1609 

## Code Review Checklist

ci:next

### DIFF
--- a/ProcessMaker/Models/ProcessRequest.php
+++ b/ProcessMaker/Models/ProcessRequest.php
@@ -1057,9 +1057,9 @@ class ProcessRequest extends ProcessMakerModel implements ExecutionInstanceInter
     /**
      * Retrieves the destination of the first closed end event.
      *
-     * @return ?string Returns a string value representing the element destination of the first closed end event.
+     * @return ?array Returns a string value representing the element destination of the first closed end event.
      */
-    public function getElementDestination(): ?string
+    public function getElementDestination(): ?array
     {
         $endEvents = $this->tokens()
             ->where('element_type', 'end_event')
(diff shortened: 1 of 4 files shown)


ProcessMaker/Models/ProcessRequest.php
```diff
@@ -1057,9 +1057,9 @@ public function getProcessVersionAlternativeAttribute(): string | null
     /**
      * Retrieves the destination of the first closed end event.
      *
-     * @return ?string Returns a string value representing the element destination of the first closed end event.
+     * @return ?array Returns a string value representing the element destination of the first closed end event.
```
caleeli, line 1060: Fix the doc description `string` -> `array`
Does this method was introduced in this feature?
Make sure to change every where this method is used to use array instead of string.

ProcessMaker/Models/ProcessRequestToken.php
```diff
@@ -136,7 +136,7 @@ class ProcessRequestToken extends ProcessMakerModel implements TokenInterface
      */
     protected $appends = [
         'advanceStatus',
-        'elementDestination'
+        'elementDestination',
```
caleeli, line 139: Do not use appends since it will execute the`getElementDestinationAttribute` for each row of the tokens causing performance problems.

ProcessMaker/Models/ProcessRequest.php
rodriquelca: Hello @caleeli yes this method was introduced in this feature.

caleeli (approved):

PR #7080, opened 2024-07-12
FOUR-17013 Icon for templates card does not have the correct size

gproly:
https://processmaker.atlassian.net/browse/FOUR-17013

Icon for templates card does not have the correct size

Normal size is 40x48 
Mobile size is 24x24

ci:next


nolanpro (approved):

gustavobascope (approved):
Approved

PR #7105, opened 2024-07-17
FOUR-17003 Fix position of buttons CANCEL and SAVE

caleeli:
## Issue & Reproduction Steps
The buttons were placed outside a </div> close.
Also there is a problem when activating the configuration tab when the page is refreshed.
 
## Solution
- Fix the position of buttons.
- Fix the activation rule of the Config tab.

## How to Test
- Open a process configuration
- Go to any tab, before the fix you will see a CANCEL and SAVE button in every tab
- With the fix the buttons are placed properly in the bottom of the config and not in every tab.
- Go to the Configuration tab
- Refresh the page 

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-17003

## Code Review Checklist

ci:next

### DIFF
--- a/resources/views/processes/edit.blade.php
+++ b/resources/views/processes/edit.blade.php
@@ -338,16 +338,17 @@
                                     </div>
                                 </div>
                             </div>
-                        </div>
-                        <div class="d-flex justify-content-end mt-2">
-                            {!! Form::button(__('Cancel'),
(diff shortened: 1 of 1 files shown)


pmPaulis:
@caleeli 
the configurations is save correctly and the buttons Save and Cancel is not showed in the other tabs, Approved!

pmPaulis (approved):

PR #7123, opened 2024-07-18
Bugfix/FOUR-16575: Bugfix/FOUR-16575: Getting OOM Memory Failures when Calling Tasks API Endpoint

agustinbusso:
**Fixes**
https://processmaker.atlassian.net/browse/FOUR-16575

ci:next

### DIFF
--- a/ProcessMaker/Console/Migration/ExtendedMigrateCommand.php
+++ b/ProcessMaker/Console/Migration/ExtendedMigrateCommand.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace ProcessMaker\Console\Migration;
+
+use Illuminate\Database\Console\Migrations\MigrateCommand as BaseMigrateCommand;
+use Illuminate\Support\Facades\Cache;
+use ProcessMaker\Models\ProcessMakerModel;
+
+class ExtendedMigrateCommand extends BaseMigrateCommand
+{
+
+    public function handle(): void
+    {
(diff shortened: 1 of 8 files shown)


tokensPM:
QA server K8S was successfully deployed https://ci-309fcb93b2.engk8s.processmaker.net

nolanpro:
@agustinbusso @caleeli I would consider using this exclude scope: https://stackoverflow.com/questions/23612221/how-to-exclude-certains-columns-while-using-eloquent/56425794#56425794

You can drop it right into ProcessMaker/Models/ProcessMakerModel.php and it should work everywhere.

Also, process_reqests also has the `data` field, and processes also has some big columns, so you should exclude them when eager loading too

```
        $query = ProcessRequestToken::exclude(['data'])->with([
            'processRequest' => fn($q) => $q->exclude(['data']),
            'process' => fn($q) => $q->exclude(['svg', 'bpmn', 'warnings']),
            'user',
            'draft'
        ]);
```

You'll need to remove `scopeExclude` from Screen.php since this replaces it.

caleeli:
The following scenarios are being tested:

- [ ] Test the Tasks table
- [ ] Test the endpoint as reported in the ticket
- [ ] Search by data.fields E.g. data.Year="2024"
- [ ] Filter by different columns
- [ ] Add data columns like data.Year
- [ ] Create a Task Saved Search that includes different columns and data fields

caleeli:
SEARCH BY DATA
image

Filter and sort by data column
image

caleeli (approved):

PR #7148, opened 2024-07-23
FOUR-17213 e.data.forEach is not a function message displayed when co…

gproly:
https://processmaker.atlassian.net/browse/FOUR-17213

e.data.forEach is not a function message displayed when completing a case

ci:next
.


eiresendez (approved):
Approved, there are no errors in the console and the filter in the collection is applied correctly.

Screenshot 2024-07-23 at 11 11 10 a m

PR #7213, opened 2024-08-01
FOUR-17536 Setup redis as default CACHE_DRIVER

gproly:
https://processmaker.atlassian.net/browse/FOUR-17536

Setup redis as default CACHE_DRIVER

In order to prevent cache errors when the enviroment is updated, we need to setup the CACHE_DRIVER by default to redis at config/cache.php

ci:next


rodriquelca (approved):

ryancooley:
@nolanpro If you can please review. I'm concerned this change might impact deployments, automated or otherwise. It looks like that may be the case as the Docker image is not building.

boliviacoca (approved):

nolanpro (approved):

PR #7219, opened 2024-08-02
Fix PM blocks launchpad import issue

nolanpro:
## Issue & Reproduction Steps
The importer automatically increments the process_id of ProcessLaunchpad because it must be unique. But the new ID was not seen by the import logic that happens later so it was attempting to save using the old (non-unique) id.

## Solution
- Run refresh() on the model before running the import logic.

## How to Test
See steps in jira issue

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-16751

## Code Review Checklist

ci:next
.


eiresendez (approved):
Approved! Works as expected!

PR #7236, opened 2024-08-08
Do not invalidate tokens when showing fallback route

nolanpro:
## Issue & Reproduction Steps
When the custom error render was using a fallback route, it was re-running it through the middleware pipeline invalidating our short-lived tokens we use for API calls (I think)

## Solution
- Bypass the middleware and just show the fallback route

## How to Test
- Open modeler with at least 2 tasks
- Click on one task
- In another tab (not incognito), open an invalid webentry url like `/webentry/9999/77777`
- Go back to modeler and without reloading, click on another task.
- You should not see any error messages

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-17525

ci:next

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Exception/Handler.php
+++ b/ProcessMaker/Exception/Handler.php
@@ -26,10 +26,10 @@ class Handler extends ExceptionHandler
      * @var array
      */
     protected $dontReport = [
-        \Illuminate\Auth\AuthenticationException::class,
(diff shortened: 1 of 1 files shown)


mcraeteisha (approved):
Approved!

PR #7286, opened 2024-08-28
FOUR-17398 Date filter is shown in ISO 8601 format and in UTC confusi…

gproly:
## Issue & Reproduction Steps
Date filters are shown in ISO 8601 format and UTC time
In the screenshot above, note that even though the date and time selected in the filter was 2024-07-26 00:00:00, the filter is showing 2024-07-26T04:00:00.000Z
This causes confusion in end users since they believe the time used are incorrect.

## Solution
Date filters should be shown in user’s date and time format and user’s timezone.
The control has been changed; it now uses the PMDatetimePicker component, which is an improved version that supports date and time formatting. Please perform a quick test of its functionality.

## Related Tickets & Packages
[- Link to any related FOUR tickets, PRDs, or packages
](https://processmaker.atlassian.net/browse/FOUR-17398)

ci:next

### DIFF
--- a/resources/js/common/advancedFilterStatusMixin.js
+++ b/resources/js/common/advancedFilterStatusMixin.js
@@ -26,7 +26,8 @@ export default {
   },
   methods: {
     setAdvancedFilter() {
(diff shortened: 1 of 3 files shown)


gustavobascope (approved):
Approved
image

PR #7356, opened 2024-09-11
FOUR-19101 Filter Tags in Tasks and Cases Pages are displaying other …

gproly:
## Issue & Reproduction Steps
An incorrect date validation method was used.

## Solution
We are using the correct validation method.

## How to Test
Check correct format.

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-19101

ci:next


estebangallego (approved):
Approved, it works as expected.

PR #7436, opened 2024-09-27
FOUR-17328 Patch generated SDK code to fix error

nolanpro:
## Issue & Reproduction Steps
This is a bug in OpenAPI Generator https://github.com/OpenAPITools/openapi-generator/issues/8943

It was apparently fixed in v6 but we are locked into 5.1.1 because, as far as I remember, in more recent versions the CSharp SDK was broken.

Even with 5.1.1 we are patching some code and this fix patches more code.

The OpenAPI Generator project has continued to have problems and we should consider depreciating its usage. We could just replace it with a pre-built guzzle client.

## Solution
- Patch the generated code.

## How to Test
- Verify you can see the problem in the jira ticket using the script editor
- Checkout this branch and rebuild the executor `php artisan processmaker:build-script-executor php --rebuild`
- Try it again and it should work

ci:next

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-17328

## Code Review Checklist

### DIFF
--- a/ProcessMaker/BuildSdk.php
+++ b/ProcessMaker/BuildSdk.php
@@ -58,7 +58,7 @@ class BuildSdk
     public function setUserId($userId)
     {
         if (!is_numeric($userId)) {
-            throw new \Exception('User id must be a number');
+            throw new Exception('User id must be a number');
         }
         $this->userId = $userId;
     }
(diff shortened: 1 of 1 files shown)


eiresendez (approved):
Tested manually! 👍  the script worked after using the patch! 

image

PR #7499, opened 2024-10-08
FOUR-19236 Encryption is Not Working in Authenticated Web Entry

julceslauhub:
https://processmaker.atlassian.net/browse/FOUR-19236

ci:next

### DIFF
--- a/ProcessMaker/Http/Controllers/Api/EncryptedDataController.php
+++ b/ProcessMaker/Http/Controllers/Api/EncryptedDataController.php
@@ -14,18 +14,19 @@ class EncryptedDataController extends Controller
     {
         // Validate request inputs
         $request->validate([
+            'uuid' => 'nullable|uuid',
             'field_name' => 'required',
-            'request_id' => 'required|exists:process_requests,id',
             'plain_text' => 'required',
             'screen_id' => 'required|exists:screens,id',
         ]);
 
         // Set variables
+        $uuid = $request->input('uuid');
         $fieldName = $request->input('field_name');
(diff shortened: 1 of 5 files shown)


ProcessMaker/Http/Controllers/Api/EncryptedDataController.php
```diff
@@ -14,18 +14,19 @@ public function encryptText(Request $request)
     {
         // Validate request inputs
         $request->validate([
+            'uuid' => 'nullable|uuid',
```
caleeli, line 17: `uuid` must be required

tests/Feature/Api/EncryptedDataTest.php
```diff
@@ -74,46 +69,17 @@ public function test_encrypt_text_field_name_empty()
         $this->assertIsArray($responseData['errors']);
     }
 
-    public function test_encrypt_text_request_id_not_found()
```
caleeli, line 77: Instead of this removed test the request_id, add a test to validate uuid is valid and is required in backend:

ProcessMaker/Http/Controllers/Api/EncryptedDataController.php
julceslauhub: in this case no, the first time will be null

tests/Feature/Api/EncryptedDataTest.php
julceslauhub: done

caleeli (approved):

PR #7540, opened 2024-10-12
FOUR-19600: The Pagination in not working in the Tabs

pmPaulis:
##Steps

1. Login in the server 
2. Go to cases list
3. Check the case # 23
4. Open the Case
5. Review the rows to return
6. 

## Current Behavior

Is only showed 10 rows

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-19600

## Code Review Checklist

### DIFF
--- a/resources/jscomposition/cases/casesDetail/config/columns.js
+++ b/resources/jscomposition/cases/casesDetail/config/columns.js
@@ -50,14 +50,6 @@ const processNameColumn = () => ({
   width: 200,
 });
 
-// const assignedColumn = () => ({
-//   field: "user.fullname",
-//   header: "Assigned",
-//   resizable: true,
-//   width: 200,
-//   filter: true,
-// });
-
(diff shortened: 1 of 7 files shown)


resources/jscomposition/cases/casesDetail/config/columns.js
```diff
@@ -109,11 +101,7 @@ const processRequestColumn = () => ({
   filter: { type: "sortable" },
   cellRenderer: () => ({
     component: LinkCell,
-    params: {
-      click: (row, column, columns) => {
-        window.document.location = `/requests/${row.id}`;
-      },
-    },
+    params: {},
   }),
```
fagubla, line 105: If the column doesn't need the Link remove all property cellRenderer

fagubla (changes requested):

fagubla (approved):

PR #7554, opened 2024-10-15
18086 Add Asset to a Bundle from the ellipsis

nolanpro:
https://processmaker.atlassian.net/browse/FOUR-18086

Also prevents editing of locked bundles


gustavobascope (approved):
Approved

PR #7655, opened 2024-10-30
FOUR-19903 Message is not displayed when there is not an instance or …

gproly:
## Message is not displayed when there is not an instance or a bundle

Go to Admin > devlink > linked instances tab
Go to Admin > devlink > local bundles tab

## Solution
- https://processmaker.atlassian.net/browse/FOUR-19903


nolanpro (approved):

gustavobascope (approved):
Approved

PR #7709, opened 2024-11-07
Observation/four 19998

gproly:
## Counters are displayed in the user list when a file is shared to other user
The users list is displaying counters on File Manager / Share

Expected Behavior:
The list should not display counters to share a file with other users. In the other hand, counters were implemented to assign users based in their assigned Tasks

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-20002


nolanpro:
Hold until https://github.com/ProcessMaker/processmaker/pull/7690 is merged

CarliPinell (approved):

PR #7769, opened 2024-11-26
FOUR-19583: Cyclical group assignment stops users from being able to login

pmPaulis:
## Issue & Reproduction Steps
##Steps to Reproduce:

Create two groups (Group A and Group B )
Assign the user you will be logging in with to Group A
Assign Group B as a member of Group A
Assign Group A as a member of Group B
Try to login with the user assigned to Group A
At this point you will see that the user can’t login, and the following screen will be shown (after waiting for a while) or an ERR_CONNECT

## Current Behavior:

When there is a cyclical group assignment as the scenario described above, users within Group A can’t login

## Expected Behavior:

Admin user has to be alerted (through an alertbox perhaps) that the group trying to be set as a member will cause cyclical assignment between groups.

When adding a group as a member to another group, we need to check if the group being assigned has the father group already there and if the user is part of it.

## Solution
- In this use case the following error is showing: `[2024-11-25 17:08:30] dev.ERROR: Allowed memory size of 134217728 bytes exhausted (tried to allocate 364544 bytes) {"userId":13,"exception":"[object] (Symfony\\Component\\ErrorHandler\\Error\\FatalError(code: 0): Allowed memory size of 134217728 bytes exhausted (tried to allocate 364544 bytes) at /Users/paulis/Herd/processmaker/ProcessMaker/Traits/HasAuthorization.php:49)`
- I validated the function to avoid the recursive when the group was verified.
- I added some unit test

## How to Test

- Create a userA assign into the groupA
- Assigne GroupB to GroupA and also GroupB to GroupA
- Login with the user A

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-19583

## Code Review Checklist
ci:deploy


tokensPM:
QA server K8S was successfully deployed https://ci-2589d41211.engk8s.processmaker.net

tokensPM:
QA server K8S was successfully deployed https://ci-2589d41211.engk8s.processmaker.net

PR #7915, opened 2025-01-17
FOUR-20562 [20302] Implement Column setup PER PROCESS for Tasks and C…

gproly:
## Issue & Reproduction Steps
[20302] Implement Column setup PER PROCESS for Tasks and Cases in Launchpad Modal window only for Designer user-part1

## Solution
- https://processmaker.atlassian.net/browse/FOUR-20562


CarliPinell (commented):
@gproly Columns assignation is working good. However there is a problem with UI when Columns Modal window is opened:
image

CarliPinell (approved):
@Functionallity is ok. Approved

PR #7925, opened 2025-01-21
Feature/FOUR-20304: Dashboards in Inbox - Implement list of dashobards asigned to Participant's group in List Pane

CarliPinell:
## How to Test
- Login PM4
- Go to Admin -> Groups
- Select some Group and EDIT
- A new dropdown called "Default Page Dashboards" should be displayed
- Choose one or more Dashboards (It is not necessary to press SAVE button)
- Go to HOME
- Select some Dashboard from left Pane

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-20304

## Code Review Checklist

### DIFF
--- a/resources/js/tasks/components/ProcessesDashboardsMenu.vue
+++ b/resources/js/tasks/components/ProcessesDashboardsMenu.vue
@@ -22,20 +22,18 @@
 
       <div class="divider-custom"></div>
 
-      <div class="dashboard-section">
-        <div class="buttons-list-dashboard">
-          <button
-            v-for="dashboard in dashboards"
-            :key="dashboard.id"
-            class="menu-btn"
-            @click="openProcessDashboard(dashboard.id, 'dashboard')"
-          >
-            <i class="fp-tachometer-alt-average"></i>
-            <span :id="`dashboard-${dashboard.id}`" class="title-dashboard">
-              {{ dashboard.name }}
-            </span>
-          </button>
-        </div>
+      <div class="buttons-list-dashboard">
+        <button
+          v-for="dashboard in dashboards"
+          :key="dashboard.id"
+          class="menu-btn"
+          @click="openProcessDashboard(dashboard.id, 'dashboard')"
+        >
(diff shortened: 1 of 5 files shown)


resources/js/tasks/components/ProcessesDashboardsMenu.vue
```diff
@@ -109,6 +121,13 @@ export default {
         &order_by=name&order_direction=asc
         &include=user,categories,category`;
     },
+    buildURLDashboards() {
+      return `/dynamic-ui/custom-dashboards`;
+    },
+    buildURLDashboardsScreen(id) {
+      //return `/dynamic-ui/custom-dashboards/${token}`;
```
nolanpro, line 128: Please remove

resources/views/tasks/index.blade.php
```diff
@@ -159,6 +159,7 @@
         window.ProcessMaker.userConfiguration = @json($userConfiguration ?? []);
         window.sessionStorage.setItem('elementDestinationURL', window.location.href);
         window.ProcessMaker.showOldTaskScreen = @json($showOldTaskScreen);
+        window.Processmaker.user = @json($currentUser);
```
nolanpro, line 162: Could you check if `window.ProcessMaker.user` has what you need?

nolanpro (changes requested):

resources/views/tasks/index.blade.php
CarliPinell: I checked that, and we have available ProcessMaker.user (with capital M) and in Screen Builder we use Processmaker.user (with lowercase m). That is the reason why I added this variable in tasks context to avoid an error when dashboard screens load in the new Participant Home screen.

resources/js/tasks/components/ProcessesDashboardsMenu.vue
CarliPinell: Comment removed.

nolanpro (approved):

PR #7926, opened 2025-01-22
FOUR-21497 - Fix script retry error handling for script microservice

nolanpro:
## Issue & Reproduction Steps
Script retries were not working with the new microservice

## Solution
- Modify the microservice-specific code to support retries

### Changed Async to Sync
In the rest-script-runner-2025-winter branch, `$sync` param is false when running async script tasks. This results in 2 async processes

`Run Script -> Async Job -> Run Microservice Script -> Run Async on Microservice`

This would complicate the Retry error handling logic and I believe it's unnecessary and could result in additional complexity in the life cycle.

So in this PR [I changed it](https://github.com/ProcessMaker/processmaker/compare/rest-script-runner-2025-winter...bugfix/FOUR-21497?expand=1#diff-08c24505c4e55a006b2fc428c4813a0bbd9c041325010f7d00fcc0cd0e5d61ecR111) to sync so the original retry logic works the same as it did previously.

## How to Test
See jira issue

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-21497

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Jobs/RunScriptTask.php
+++ b/ProcessMaker/Jobs/RunScriptTask.php
@@ -108,7 +108,7 @@ class RunScriptTask extends BpmnAction implements ShouldQueue
                     'attempts' => $this->attemptNum,
                 ],
             ];
-            $response = $script->runScript($data, $configuration, $token->getId(), $errorHandling->timeout(), 0, $metadata);
+            $response = $script->runScript($data, $configuration, $token->getId(), $errorHandling->timeout(), 1, $metadata);
 
             if (!config('script-runner-microservice.enabled') ||
                 ($scriptExecutor && $scriptExecutor->type === ScriptExecutorType::Custom)) {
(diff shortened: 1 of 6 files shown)


ProcessMaker/Jobs/RunScriptTask.php
```diff
@@ -108,7 +108,7 @@ public function action(ProcessRequestToken $token = null, ScriptTaskInterface $e
                     'attempts' => $this->attemptNum,
                 ],
             ];
-            $response = $script->runScript($data, $configuration, $token->getId(), $errorHandling->timeout(), 0, $metadata);
+            $response = $script->runScript($data, $configuration, $token->getId(), $errorHandling->timeout(), 1, $metadata);
```
epdl, line 111: @nolanpro here we are forcing this to be sync. This is not the idea, we should control this in the callback response.

epdl (changes requested):

ProcessMaker/Jobs/RunScriptTask.php
nolanpro: @epdl in that case, should we remove ProcessMaker/Jobs/RunScriptTask.php and make this entirely synchronous from core's point of view? Do you see any point in using horizon/queues in core for scripts? Queueing one job that queues another job would make the process error prone

nolanpro:
Closing in favor of https://github.com/ProcessMaker/processmaker/pull/7949

PR #8041, opened 2025-02-17
bugfix/FOUR-21978 update the counter when the process info is collapsed

gustavobascope:
## Issue & Reproduction Steps
The counter is not update when the Process info collapses

## Solution
update the counter when the process info is collapsed

## How to Test
Create a process 
Open the process by launchpad
Start a case
Collapse the process info
Check the counter in the collapsible process info and the counter 

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-21978


marcoAntonioNina (approved):
Works as expected

PR #8094, opened 2025-03-18
Observation/FOUR-23367: Deleted dashboards continue to display on Home screen

CarliPinell:
##IMPORTANT
- Both Branches CORE and Package-dymanic-ui MUST be deployed.

## Solution
- New logic was added to remove dashboards from custom selected dashboards for New Participant Home Screen
- New logic was added to check previous Custom Dashboards assigned to Users or Groups

## How to Test
- Login PM
- Create multiple users and assign to different Groups
- Create Multiple Dashboards
- Configure several use cases with Default Home Screens, Dashboards, Analytics, URL, TOP-MENU
- Assign all previous settings to Users and groups
- Login and Logout with different users.

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-23367

## Code Review Checklist


ProcessMaker/Http/Controllers/HomeController.php
```diff
@@ -15,12 +16,27 @@ public function index(Request $request)
...
                     ->where('assignable_id', $user->id)
+                    ->where('assignable_type', 'ProcessMaker\Models\User')
                     ->count() > 0;
 
+                //Check if there is at least one custom dashboard per group only first group is selected
+                if (!$customDashboardExists) {
+                   $firstGroupId = Arr::first($groups);
+                    $customDashboardExists = \ProcessMaker\Package\PackageDynamicUI\Models\DynamicUI::where('type', 'DASHBOARD')
+                        ->where('assignable_type', 'ProcessMaker\Models\Group')
+                        ->where('assignable_id', $firstGroupId)
```
nolanpro, line 36: @CarliPinell should this be `->whereIn('assignable_id', $groups)` because who knows if the first group has any dashboards? Maybe another group does.

nolanpro (changes requested):

ProcessMaker/Http/Controllers/HomeController.php
CarliPinell: I talked with Alejandra about that, and she told me that in case of one user is assigned to more than one group, only the first group ID founded should be considered. That is the reason we don't make a search with an array with more than one Group Id, because we only can render one screen.

nolanpro (approved):

PR #8114, opened 2025-03-21
FOUR-22988 - Fix export when nested screen does not exist

nolanpro:
Fix export when nested screen does not exist

See https://processmaker.atlassian.net/browse/FOUR-22988


gustavobascope (approved):
Approved

PR #8151, opened 2025-03-27
FOUR-23643 Fix migration issue

caleeli:
## Issue & Reproduction Steps
When the migration ran, there was a record in  dev_links with client_secret in null.

image

## Solution
- Fix the migration that convert the column to text.

image

## How to Test
Migrate an environment with a dev link configured without client_secret (null).

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-23643

## Code Review Checklist

### DIFF
--- a/database/migrations/2025_02_26_194048_modify_client_secret_in_dev_links_table.php
+++ b/database/migrations/2025_02_26_194048_modify_client_secret_in_dev_links_table.php
@@ -11,7 +11,7 @@ return new class extends Migration {
     public function up(): void
     {
         Schema::table('dev_links', function (Blueprint $table) {
-            $table->text('client_secret')->change();
+            $table->text('client_secret')->nullable()->change();
         });
     }
 
@@ -21,7 +21,7 @@ return new class extends Migration {
     public function down(): void
     {
         Schema::table('dev_links', function (Blueprint $table) {
-            $table->string('client_secret')->change();
+            $table->string('client_secret')->nullable()->change();
         });
     }
 };


ryancooley (commented):
@caleeli We're sure it's okay to modify this migration instead of creating a new one? Modifying migrations has had unintended consequences in the past.

caleeli:
@ryancooley yes, it is required, since that migration was causing the problem.

## Without fix:
### Case 1. Fresh installation ✅
- modify_client_secret_in_dev_links_table runs without problem (there is no null values)
- make_dev_links_client_secret_nullable also runs (adds the nullable property)

### Case 2. Update (without null values) ✅
- modify_client_secret_in_dev_links_table runs without problem (there is no null values)
- make_dev_links_client_secret_nullable also runs (adds the nullable property)

### Case 3. Update (with null values) ❌
- modify_client_secret_in_dev_links_table [❌ FAILS because null value]
- make_dev_links_client_secret_nullable also runs (does not do any change)

## With fix:

### Case 1. Fresh installation ✅
- modify_client_secret_in_dev_links_table runs without problem (adds the nullable property)
- make_dev_links_client_secret_nullable also runs (does not do any change)

### Case 2. Update (without null values) ✅
- modify_client_secret_in_dev_links_table runs without problem (adds the nullable property)
- make_dev_links_client_secret_nullable also runs (does not do any change)

### Case 3. Update (with null values) ✅
- modify_client_secret_in_dev_links_table runs without problem (adds the nullable property)
- make_dev_links_client_secret_nullable also runs (does not do any change)

There are two migrations that do the same, in case one was alredy executed (withut the fix), the second introduces the fix.

PR #8162, opened 2025-04-01
observation/FOUR-23694 when settings configurations were added without any asset in the bundle then reinstall or update are not working.

gustavobascope:
## Issue & Reproduction Steps
when settings configurations were added without any asset in the bundle then reinstall or update are not working.

## Solution
fix the reinstall bundle feature when the bundle only has settings

## How to Test
Create a bundle without assets and settings and publish
on the other instance install the bundle
after installing go to the bundle and reinstall the bundle

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-23694


CarliPinell (approved):
approved

PR #8184, opened 2025-04-10
FOUR-23830 - Remove link

nolanpro:
See https://processmaker.atlassian.net/browse/FOUR-23830

Requires https://github.com/ProcessMaker/pm4-k8s-distribution/pull/282

### DIFF
--- a/upgrades/2025_04_10_162213_remove_storage_symlink.php
+++ b/upgrades/2025_04_10_162213_remove_storage_symlink.php
@@ -1,0 +1,47 @@
+<?php
+
+use Illuminate\Support\Facades\Artisan;
+use ProcessMaker\Upgrades\UpgradeMigration as Upgrade;
+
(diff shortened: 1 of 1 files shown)


upgrades/2025_04_10_162213_remove_storage_symlink.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Run the upgrade migration.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Artisan::call('storage:unlink');
```
caleeli, line 35: This breaks the access to some publc files like profile avatar

image

nolanpro:
Closing because cloudops added a temporary fix in nginx and this fix would have blocked profile pics like David mentioned. This will need to be revisited.

PR #8209, opened 2025-04-25
FOUR-23473 S1: Add stage section for a the flow elements

gproly:
##Description
Add Stage Section showed only when the element selected is a FLOW

https://processmaker.atlassian.net/browse/FOUR-23473


pmPaulis (approved):

pmPaulis:
Approved, we can see some empty templated but I know that is the initial PR about the changes

PR #8338, opened 2025-07-02
FOUR-25078 Response time in data connector resources has increased afer the upgrade to version 4.14

julceslauhub:
## Issue & Reproduction Steps
Endpoint to get all records from a collection is very slow

## Solution
Improvement in the local cache for settings

## How to Test
Use the endpoint: 
`http://<server>/api/1.0/collections/<collection_id>/records`
with a collection with 1500 records at least

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-25078

## Code Review Checklist

### DIFF
--- a/tests/Repositories/SettingsConfigRepositoryTest.php
+++ b/tests/Repositories/SettingsConfigRepositoryTest.php
@@ -20,4 +20,23 @@ class SettingsConfigRepositoryTest extends TestCase
 
         $this->assertEquals('the value', config('test.dot.notation'));
     }
+
+    public function testCachesValueSimple()
+    {
+        Setting::create([
+            'key' => 'cache-test',
+            'config' => 'foo',
+            'format' => 'text',
+        ]);
+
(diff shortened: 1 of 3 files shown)


caleeli (changes requested):
@julceslauhub please add some tests to validate the fix.

nolanpro (approved):

tests/Repositories/SettingsConfigRepositoryTest.php
```diff
@@ -20,4 +20,23 @@ public function testDotNotation()
 
         $this->assertEquals('the value', config('test.dot.notation'));
     }
+
+    public function testCachesValueSimple()
+    {
+        Setting::create([
+            'key' => 'cache-test',
```
caleeli, line 27: To keep reference with the Collections code could you change `'cache-test'` by smoething like `'collections.6.obfuscate'` and add a comment to reference the class were it is used.

PR #8398, opened 2025-07-29
FOUR-25187: Processes cannot be created with BPMN files

julceslauhub:
## Issue & Reproduction Steps
The BPMN files generated by the modeler used by the client cannot be imported into PM. However, it works perfectly with other BPMN files from different modelers.

## Solution
The XSD definition used by ProcessMaker restricts some definitions found in the client’s BPMN files. A couple of these restrictions have been modified to allow the client’s files to be imported.

## How to Test
Create a new process and use one of the BPMN files provided in the ticket.

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-25187

## Important Note

With this change, the BPMN files provided by the client can be imported. However, our modeler does not automatically fix the visual layout of certain elements in their designs (likely due to differences in scaling and icon sizes between the two modelers).

## Code Review Checklist

### DIFF
--- a/ProcessMaker/Http/Controllers/Api/ProcessController.php
+++ b/ProcessMaker/Http/Controllers/Api/ProcessController.php
@@ -383,7 +383,7 @@ class ProcessController extends Controller
 
             // Some BPMN definitions created in others modelers doesn't comply BPMN 2.0
             // So, should be fixed
-            $fixBpmnSchemaService = new FixBpmnSchemaService();
+            $fixBpmnSchemaService = app(FixBpmnSchemaService::class);
             $data['bpmn'] = $fixBpmnSchemaService->fix($data['bpmn']);
 
             $request->merge(['bpmn' => $data['bpmn']]);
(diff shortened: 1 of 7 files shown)


caleeli (commented):
Please review is the Semantic.xsd matches the official standar https://www.omg.org/spec/BPMN/20100501/Semantic.xsd

ProcessMaker/Http/Controllers/Api/ProcessController.php
```diff
@@ -379,6 +380,12 @@ public function store(Request $request)
         // Validate if exists file bpmn
         if ($request->has('file')) {
             $data['bpmn'] = $request->file('file')->get();
+
+            // Some BPMN definitions created in others modelers doesn't comply BPMN 2.0
+            // So, should be fixed
+            $fixBpmnSchemaService = new FixBpmnSchemaService();
```
caleeli, line 386: Register the service and use the Laravel style to register services:

```
$fixBpmnSchemaService = app(FixBpmnSchemaService::class);
```

julceslauhub: Done.

caleeli (changes requested):
Please add a test to the implementation

caleeli (approved):

PR #8474, opened 2025-08-25
FOUR-23421 There is confusion with the stages when they are not publi…

gproly:
…shed.

 ## Description:
Create a process
Add Start Event → Task Form → End Event
Create two stages and assigned these to flow to each task Publish the process
Open the process by launchpad
Create a case
Return to process
Delete the stages in the process
Does not publish the changes
Go to process launchpad
Check that stages are not present after the changes Create new cases
The new cases are created with stage but it is not visible in process launchpad Current Behavior:
To stages are visible to start a CASE is necessary to publish the process with stages To stages are visible in LAUNCH PAD does not necessary to publish the process with stages which causes confusion when displaying case information Expected Behavior:
The information in launch pad should be sync
If there are stages the new cases should be created with the corresponding stage If there are not stages the new cases should be created without stages To this it necessary that stages will be publish to avoid errors and confusions

 ## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-25926

ci:deploy

### DIFF
--- a/resources/js/processes/modeler/components/inspector/StageManager.vue
+++ b/resources/js/processes/modeler/components/inspector/StageManager.vue
@@ -73,6 +73,10 @@ const getHighlightedNode = () => getModeler().highlightedNode;
 
 const getDefinition = () => getHighlightedNode().definition;
 
+const saveProcess = () => {
+  currentInstance.proxy.$root.$children[0].$refs["external-ModalSaveVersion"]?.[0]?.saveModal?.();
+}
+
 const getConfigFromDefinition = (definition) => {
   let config = {};
   try {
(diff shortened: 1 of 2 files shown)


resources/js/processes/modeler/components/inspector/StageManager.vue
```diff
@@ -73,6 +73,10 @@ const getHighlightedNode = () => getModeler().highlightedNode;
 
 const getDefinition = () => getHighlightedNode().definition;
 
+const saveProcess = () => {
+  currentInstance.proxy.$root.$children[0].$refs["external-ModalSaveVersion"]?.[0]?.saveModal?.();
```
fagubla: 0 is hardcode

fagubla (changes requested):

vladyrichter:
QA server K8S was successfully deployed https://ci-7506e8d01a.engk8s.processmaker.net

pmPaulis (approved):

PR #8489, opened 2025-08-29
Fix bundle export when user is null

nolanpro:
See https://processmaker.atlassian.net/browse/FOUR-26146


gustavobascope (approved):
Approved

PR #8540, opened 2025-09-26
FOUR-26712 Implement the Events and Logic in Backed for Conditional Destination

caleeli:
## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-26712


ProcessMaker/Models/ProcessRequestToken.php
```diff
@@ -1308,10 +1309,21 @@ public function reassign($toUserId, User $requestingUser)
      *
      * @return array|null Returns the destination URL.
      */
-    private function getElementDestination($elementDestinationType, $elementDestinationProp): ?array
+    private function getElementDestination($elementDestinationType, $elementDestinationProp, array $conditionalRedirectProp): ?array
     {
         $elementDestination = null;
 
+        if (count($conditionalRedirectProp) > 0) {
+            $result = $this->evaluateConditionalRedirect(app(ConditionalRedirectServiceInterface::class), $conditionalRedirectProp);
```
rodriquelca, line 1317: maybe will be need validate if $conditionalRedirectProp['isEnabled'] and if there is conditions property

rodriquelca (approved):

ProcessMaker/Services/ConditionalRedirectService.php
```diff
@@ -0,0 +1,208 @@
...
+        foreach ($conditionalRedirect as $item) {
+            if (!isset($item['condition'])) {
+                throw new InvalidArgumentException('Condition is required');
+            }
+
+            $condition = $item['condition'];
+
+            $this->feel->setBody($condition);
+            try {
+                $result = ($this->feel)($data);
```
rodriquelca, line 116: a new instance is created for everi conditions, if we have 10 then then instances will be created

ProcessMaker/Services/ConditionalRedirectService.php
```diff
@@ -0,0 +1,208 @@
...
+            foreach ($this->errors as $error) {
+                $this->addLogComment($token, $error);
+            }
+        }
+        return $result;
+    }
+
+    private function addLogComment(ProcessRequestToken $token, string $error)
+    {
+        $case_number = ProcessRequest::where('id', $token->process_request_id)->value('case_number');
```
rodriquelca: is posible use  to   `$case_number = $token->processRequest->case_number` to prevent query n +1 ?

ProcessMaker/Models/ProcessRequestToken.php
caleeli: this condition is being validated inside `private function evaluateConditionalRedirect`

ProcessMaker/Services/ConditionalRedirectService.php
caleeli: `$this->feel` callable is created once in the constructor, at this line to get the reference of the callable I had to use `($this->feel)` to avoid confution with normal functions, but it is not creating a new instance for each condition.

ProcessMaker/Services/ConditionalRedirectService.php
caleeli: Great catch, code was improved to check if the relationship was already loaded and moved outside the loop to avoid n+1 querires

ProcessMaker/Models/ProcessRequestToken.php
caleeli: code was improved

rodriquelca (approved):

PR #8578, opened 2025-10-29
Fix scheduler when tenants have different db credentials

nolanpro:
ci:deploy
ci:multitenancy

https://processmaker.atlassian.net/browse/FOUR-27667


gusys (approved):
Looks good to me

PR #8649, opened 2025-12-12
FOUR-24108 Reassignment form is not hidden in tasks where the "Allow …

gproly:
…Reassignment" option is not enabled.

 ## Description
Reassignment form is not hidden in tasks where the "Allow Reassignment" option is not enabled. Expected Behavior:
When the task is changed by next button . the reassignment form should verify if this options is enable for the task If the option is not enable the reassign form should hidden
 ## Related tickets
https://processmaker.atlassian.net/browse/FOUR-24108

 ## Solution
There are two sections of the interface that consume the same endpoint. This endpoint is designed to support both GET and POST. A change was made to use POST in order to send FormData. However, a problem was found: when search strings were introduced, the interface continued using the GET method and did not send FormData. Therefore, the data-retrieval logic for this endpoint is centralized into a single function to ensure consistent behavior across both interfaces.

Additionally, permissions are not required to maintain compatibility with the previous version, because the endpoint was originally created with permissions in a way that did not account for backward compatibility with the previous behavior.

ci:deploy


vladyrichter:
QA server K8S was successfully deployed https://ci-9a90e3b0a8.engk8s.processmaker.net

fagubla (approved):

ana-mauricio (approved):

vladyrichter:
QA server K8S was successfully deployed https://ci-9a90e3b0a8.engk8s.processmaker.net

PR #8713, opened 2026-01-30
FOUR-29068 Add NAYRA_REST_API_HOST support to ScriptDockerNayraTrait

gproly:
When NAYRA_REST_API_HOST is set in .env, the trait uses that URL directly instead of Docker discovery, enabling Nayra script execution where automatic discovery fails (macOS/Docker Desktop, port conflicts, external Nayra services).

Problem (without this change):
- ScriptDockerNayraTrait uses Docker discovery (docker inspect, hostname -i) to build http://{ip}:{port}/run_script
- On macOS/Docker Desktop: IPv6 from hostname -i produces invalid URLs; container IP may be unreachable from host
- Port 8080 conflict: nginx or other service responds instead of Nayra
- External Nayra: no way to configure a fixed URL

Solution:
- Add getNayraBaseUrl(): returns NAYRA_REST_API_HOST if set, else getNayraInstanceUrl()
- Update handleNayraDocker() to use getNayraBaseUrl()
- Update ensureNayraServerIsRunning(): when NAYRA_REST_API_HOST is set and connection fails, throw ScriptException instead of bringUpNayra()

Configuration (.env):
  NAYRA_REST_API_HOST=http://localhost:8081

Backward compatible: if not set, original Docker discovery flow is used.

Affected file: ProcessMaker/Models/ScriptDockerNayraTrait.php

Related tickets:
https://processmaker.atlassian.net/browse/FOUR-29068

ci:deploy


vladyrichter:
QA server K8S was successfully deployed https://ci-d774665dfc.engk8s.processmaker.net

pmPaulis (approved):

PR #8781, opened 2026-04-17
FOUR-30703 | Save Search Shows “Error Loading Items” in Available Columns

mcraeteisha:
# Issue
Ticket: [FOUR-30703](https://processmaker.atlassian.net/browse/FOUR-30703)

When hitting `/api/1.0/saved-searches/{id}/columns?include=available,data,default`, the app returned an error:

> Column 'id' in EXISTS subquery is ambiguous

This only happened when a text search filter (e.g. "smart extract") was applied to a task-based Saved Search.

# Root Cause
The Saved Search package recently changed how it fetches task data (commit `9ee8acc`, ticket [FOUR-28867](https://processmaker.atlassian.net/browse/FOUR-28867)). To fix a CPU usage problem, it switched from a simple query to one that looks at the `process_request_tokens` table twice and stitches the results together side by side.

The issue is that both copies of the table have columns named `id` and `updated_at`. When the core `scopeFilter` method in `ProcessRequestToken` said "filter where `id` matches the search term," the database couldn't tell which `id` was meant — the one from the original table or the one from the copy.

# Solution
In `ProcessMaker/Models/ProcessRequestToken.php`, the `scopeFilter` method now uses fully qualified column names so the database always knows which table is being referenced:
| Before | After |
|---|---|
| `'id'` | `'process_request_tokens.id'` |
| `'created_at'` | `'process_request_tokens.created_at'` |
| `'due_at'` | `'process_request_tokens.due_at'` |
| `'updated_at'` | `'process_request_tokens.updated_at'` |

Added two tests in `tests/Feature/Api/TasksTest.php`:
- **`testScopeFilterWorksWithJoinedSubquery`** — Reproduces the exact bug scenario by building a query with a joined subquery (mimicking the Saved Search CTE pattern), then applying a text filter. Before the fix, this threw the ambiguous column error. Now it passes.
- **`testFilterByFreeTextSearch`** — Exercises the free-text `filter` parameter through the tasks API endpoint to confirm end-to-end filtering works.

# How To Test
1. Run `phpunit tests/Feature/Api/TasksTest.php`
    - Verify `testScopeFilterWorksWithJoinedSubquery` passes
    - Verify `testFilterByFreeTextSearch` passes
2. Log in to ProcessMaker. Start a couple of Cases.
3. Go to Tasks and filter by the Process name via the search bar.
4. Create a Saved Search based on the search results.
5. Go to Configure → Columns.
6. Available columns should be listed; the "Error loading items" error should not display.

ci:deploy

# Code Review Checklist
- [ ]  This code adheres to [ProcessMaker Coding Guidelines](https://github.com/ProcessMaker/processmaker/wiki/Coding-Guidelines).

[FOUR-30703]: https://processmaker.atlassian.net/browse/FOUR-30703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FOUR-28867]: https://processmaker.atlassian.net/browse/FOUR-28867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ProcessMaker/Models/ProcessRequestToken.php
+++ b/ProcessMaker/Models/ProcessRequestToken.php
@@ -651,10 +651,10 @@ class ProcessRequestToken extends ProcessMakerModel implements TokenInterface
                 $query->where(DB::raw('LOWER(element_name)'), 'like', $filter)
                     ->orWhere(DB::raw('LOWER(data)'), 'like', $filter)
                     ->orWhere(DB::raw('LOWER(status)'), 'like', $filter)
-                    ->orWhere('id', 'like', $filter)
-                    ->orWhere('created_at', 'like', $filter)
-                    ->orWhere('due_at', 'like', $filter)
-                    ->orWhere('updated_at', 'like', $filter)
+                    ->orWhere('process_request_tokens.id', 'like', $filter)
+                    ->orWhere('process_request_tokens.created_at', 'like', $filter)
+                    ->orWhere('process_request_tokens.due_at', 'like', $filter)
+                    ->orWhere('process_request_tokens.updated_at', 'like', $filter)
                     ->orWhereHas('processRequest', function ($query) use ($filter) {
                         $query->where(DB::raw('LOWER(name)'), 'like', $filter);
                     })
--- a/ProcessMaker/Models/ProcessRequestToken.php
+++ b/ProcessMaker/Models/ProcessRequestToken.php
@@ -640,10 +640,10 @@ class ProcessRequestToken extends ProcessMakerModel implements TokenInterface
         $setting = Setting::byKey('indexed-search');
         if ($setting && $setting->config['enabled'] === true) {
             if (is_numeric($filter)) {
-                $query->whereIn('id', [$filter]);
+                $query->whereIn('process_request_tokens.id', [$filter]);
             } else {
                 $matches = self::search($filter)->take(10000)->get()->pluck('id');
-                $query->whereIn('id', $matches);
+                $query->whereIn('process_request_tokens.id', $matches);
             }
         } else {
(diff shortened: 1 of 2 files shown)


Kookster310:
QA server K8S was successfully deployed https://ci-08a5fe0536.engk8s.processmaker.net

eiresendez (approved):
The PR does fix the reported issue in `FOUR-30703` for the non-indexed search path. I was able to confirm that the original Saved Search columns request no longer fails in that scenario.

There is still an open issue when `indexed-search` is enabled. In that path, the query still uses an unqualified `id`, and the Saved Search columns endpoint can still fail with `Column 'id' in EXISTS subquery is ambiguous`.

I reproduced the remaining issue locally with:

```bash
curl -sk 'https://processmaker.test/api/1.0/saved-searches/28/columns?include=available,data,default' \
  -H 'Accept: application/json' \
  -H 'Authorization: Bearer <TOKEN>'
```

Response:

```text
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in EXISTS subquery is ambiguous
```

Approved the current fix as resolving the original bug in the default path, but please keep a follow-up issue open for the `indexed-search` branch of `ProcessRequestToken::scopeFilter()`, since that path still appears to be incomplete.

Kookster310:
QA server K8S was successfully deployed https://ci-08a5fe0536.engk8s.processmaker.net

eiresendez (approved):
👍 🎉

PR #8782, opened 2026-04-20
bugfix/FOUR-26348: The status filter doesn't work when filtering two statuses.

CarliPinell:
## Issue & Reproduction Steps
After applying 2 filters to status column , the filter displays the message "No results have been found."

## Solution
-Task inbox advanced filters: Group filter rows by column even when _column_field is missing or "N/A", by resolving status and process_version_alternative from the filter subject (resolveAdvancedFilterGroupKey + isTaskStatusColumnFilter). This lets multiple status rows merge correctly instead of being split across buckets and AND-ed by the API.

-Multiple = on the same enum column: Instead of collapsing to operator: "in", multiple = values for status / process_version_alternative are serialized as a single filter object with nested or chains of operator: "=" (same OR semantics as IN, compatible with ProcessMaker\Filters\Filter).

Default inbox status: Default “In Progress” (and URL-driven variants) uses the same subject shape as the column filter (Field + status) so it merges cleanly with user-defined status filters.

Backward compatibility

API contract: Payload remains a JSON array of advanced-filter objects consumed by the existing Filter::filter() pipeline; or nesting is already supported (see existing filter tests). Single-value filters and non-enum columns are unchanged.

Behavior: One status still behaves as before; multiple status values still mean OR (tasks matching any selected state), not AND—only the representation changed from in to nested = + or.
Saved searches / stored configs: Filters that already had _column_field behave as before; configs that lacked it now group correctly instead of failing silently.

## How to Test
- Login PM4
- Go to Tasks
- In Status Column add criteria ="In Progress" + ="Completed" and Apply
- Both status must be shown

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-26348

ci:deploy

## Code Review Checklist

### DIFF
--- a/resources/js/common/PMColumnFilterPopoverCommonMixin.js
+++ b/resources/js/common/PMColumnFilterPopoverCommonMixin.js
@@ -1,4 +1,5 @@
 import { get, cloneDeep } from "lodash";
+import { isTaskStatusColumnFilter } from "./setDefaultAdvancedFilterStatus";
 
 const PMColumnFilterCommonMixin = {
   props: {
@@ -37,18 +38,66 @@ const PMColumnFilterCommonMixin = {
     });
   },
   methods: {
-    storeFilterConfiguration() {
-      const {order, type} = this.filterConfiguration();
+    /**
+     * Task inbox: if no Status column filter is present, default to the same behavior as
+     * tasks/index.js (In Progress, or Completed / Self Service from URL query).
+     */
+    applyTaskInboxDefaultStatusIfMissing() {
+      if (typeof this.filterConfiguration !== "function") {
+        return;
+      }
+      if (this.filterConfiguration().type !== "taskFilter") {
+        return;
+      }
+      if (this.advancedFilterProp !== null) {
+        return;
+      }
+      const savedSearch = this.$props.savedSearch;
+      if (savedSearch !== false && savedSearch != null && savedSearch !== "") {
+        return;
+      }
+      this.advancedFilterInit();
+      const statusFilters = this.advancedFilter.status;
+      if (Array.isArray(statusFilters) && statusFilters.length > 0) {
+        return;
+      }
 
+      let value = "In Progress";
+      try {
+        const params = new URL(document.location).searchParams;
+        if (params.get("status") === "CLOSED") {
+          value = "Completed";
+        } else if (params.get("status") === "SELF_SERVICE") {
+          value = "Self Service";
+        }
+      } catch (e) {
+        // no window (tests)
+      }
+
+      this.advancedFilter.status = [
+        {
+          subject: { type: "Field", value: "status" },
+          operator: "=",
+          value,
+          _column_field: "status",
+          _column_label: "Status",
+        },
+      ];
+    },
+    storeFilterConfiguration() {
       // If advanced filter was provided as a prop, do not save the filter
       // or overwrite the global advanced_filter, instead emit the filter.
       if (this.advancedFilterProp !== null) {
+        const { order } = this.filterConfiguration();
         this.$emit("advanced-filter-updated", {
           filters: this.formattedFilter(),
           order
         });
         return;
       }
+
+      this.applyTaskInboxDefaultStatusIfMissing();
+      const { order, type } = this.filterConfiguration();
 
       let url = "users/store_filter_configuration/";
       if (this.$props.columns && this.savedSearch) {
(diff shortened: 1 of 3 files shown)


resources/js/common/PMColumnFilterPopoverCommonMixin.js
```diff
@@ -151,9 +200,99 @@ const PMColumnFilterCommonMixin = {
...
+     * Columns that use stringSelect with "=" rows; multiple sibling "=" filters are AND-ed
+     * by the API and can never match (e.g. status In Progress AND Completed). Collapse those
+     * into one filter object using nested `or` so each value stays operator "=" (OR semantics
+     * in SQL — same as IN). Leaves existing OR-nested structures unchanged.
+     *
+     * @param {Array} filters - Filters for one column from PMColumnFilterForm
+     * @param {string} columnField - Column field name (e.g. status)
+     * @returns {Array}
+     */
+    mergeFlatEnumEqualsToOrChainForColumn(filters, columnField) {
```
eiresendez, line 232: 🔴 Critical Issues 
This PR changes explicit AND semantics into OR for multiple = filters on status / process_version_alternative.

Impact: if a user adds Status = In Progress AND Status = Completed, returning no results is expected because a single row cannot have both statuses. 

The PR silently rewrites that into OR behavior, which contradicts the selected logical operator and may change valid filter semantics.

Suggested fix: do not merge this until Product/QA confirms the desired UX. If multiple status values should mean OR, the UI should make that explicit by forcing/showing OR, using in, or preventing invalid AND combinations.

eiresendez (changes requested):
this may not be a bug. If QA used AND, “No results have been found” is expected behavior. The Jira ticket currently says “filter two statuses” but does not specify whether the logical operator was AND or OR, so with this PR we are making a product decision in code.

I would treat this as needing Product/QA clarification before merge. The safer resolution may be to update Jira and close/rework the PR rather than changing filter semantics.

@CarliPinell  what do you think?

resources/js/common/PMColumnFilterPopoverCommonMixin.js
CarliPinell: solved

resources/js/common/PMColumnFilterPopoverCommonMixin.js
```diff
@@ -37,19 +38,67 @@ const PMColumnFilterCommonMixin = {
...
       if (this.advancedFilterProp !== null) {
+        const { order } = this.filterConfiguration();
         this.$emit("advanced-filter-updated", {
           filters: this.formattedFilter(),
           order
         });
         return;
       }
 
+      this.applyTaskInboxDefaultStatusIfMissing();
```
eiresendez: This re-adds the default task Status filter before every save.

**Impact:** clearing the Status column in /tasks will immediately reinsert In Progress, because onClear() empties the column and then storeFilterConfiguration() calls applyTaskInboxDefaultStatusIfMissing(). That changes a basic user action outside the ticket scope.

**Suggested fix:** only apply the default status during initial inbox setup, not during every store/clear path.

resources/js/components/PMColumnFilterPopover/PMColumnFilterForm.vue
```diff
@@ -236,7 +243,32 @@
         }
         return operators;
       },
-      getLogicals() {
+      /**
+       * stringSelect columns (status, process version) only support OR between "=" rows;
+       * a single row cannot satisfy two different enum values with AND.
+       */
+      requiresOrLogicalOperator(index) {
```
eiresendez, line 250: This is encoding a **Product decision:** multiple `=` filters on enum-like columns become OR.

**Impact**: `Status = In Progress AND Status = Completed` returning no results is expected boolean behavior. The July 3 update makes OR visible for new UI entries, which is better, but legacy/persisted flat rows are still rewritten to OR on load/save.

**Suggested fix:** do not merge until Product/QA confirms the intended UX. If they want multi-status to mean OR, Jira should say that explicitly.

eiresendez (changes requested):
I want to flag that I do not currently see this as a confirmed bug. 

The original Jira report is ambiguous and appears to treat expected AND behavior as a defect: `Status = In Progress AND Status = Completed` should return no results because one task/request cannot have both statuses at the same time. The ticket’s expected behavior says either to “show filtered tasks/requests” or hide the + button, but those are different product choices. 

This PR changes the filter semantics and UI behavior to match what the QA reporter seems to expect, so I think Product/QA should confirm the intended behavior before we treat this as a bug fix.

cc: @sanjacornelius @CarliPinell

resources/js/common/PMColumnFilterPopoverCommonMixin.js
```diff
@@ -151,8 +200,99 @@ const PMColumnFilterCommonMixin = {
...
+      const subjectFor = (row) =>
+        row.subject && (row.subject.value !== undefined || row.subject.type === "Status")
+          ? { ...row.subject }
+          : defaultSubject;
+      const equalsLeaf = (row) => ({
+        subject: subjectFor(row),
+        operator: "=",
+        value: row.value,
+      });
+      let node = equalsLeaf(uniqueRows[uniqueRows.length - 1]);
```
eiresendez, line 275: Multi-status values are now serialized as nested `or` filters, but the task backend only detects `Self Service` from top-level status filter values. `advancedFilterHasSelfServiceStatus()` and the process-manager branch in `TaskControllerIndexMethods` do not traverse nested `or` filters, so combinations like `In Progress OR Self Service` can fail to remove the PMQL `user_id` constraint or miss process-manager self-service handling.  

**Impact**: Self Service tasks can disappear when combined with another status, especially with PMQL or `processesIManage=true`.  

**Suggested fix**: either keep multi-status as `operator: "in"` with a value array, or update backend status extraction to recursively flatten nested `or` status filters before Self Service detection/preprocessing.

resources/js/common/PMColumnFilterPopoverCommonMixin.js
eiresendez: 🔴 Critical Issues

- `[resources/js/common/PMColumnFilterPopoverCommonMixin.js:99]` `storeFilterConfiguration()` re-applies the default inbox Status whenever the status bucket is empty. `onClear()` sets `this.advancedFilter[index] = []`, then immediately calls `storeFilterConfiguration()`, which inserts `In Progress`/URL status again before persisting and fetching.  
  **Impact**: users can no longer clear the Status column filter in the task inbox for the current fetch; the filter is silently restored.  
  **Suggested fix**: apply the default only during initial task-list bootstrap, not during user-initiated store/clear operations, or explicitly distinguish “missing initial default” from “user cleared Status”.

resources/js/common/PMColumnFilterPopoverCommonMixin.js
```diff
@@ -151,8 +200,99 @@ const PMColumnFilterCommonMixin = {
...
+          continue;
+        }
+        seen.add(f.value);
+        uniqueRows.push(f);
+      }
+      if (uniqueRows.length < 2) {
+        return uniqueRows;
+      }
+      const subjectSource =
+        filters.find((f) => f.subject?.value === columnField)
```
eiresendez, line 260: 🟠 Improvements (should fix)

- `[resources/js/components/PMColumnFilterPopover/PMColumnFilterForm.vue:258]` `normalizeEnumEqualsLogicals()` only forces `or`; it never restores a previously forced `or` when the pair stops being two `=` rows.  
  Reasoning: after changing one row to `in` or removing/reordering rows, stale `or` state can survive even though the UI again allows `and`. This is lower risk than the two blockers, but it can produce surprising filter payloads.

eiresendez (changes requested):
- Since Product has accepted the behavior change, I did not treat AND to OR as a product concern. The remaining concern is that the PR changes the payload shape in a way backend special cases do not fully understand.
- Existing backend tests cover Status with array values, but not the nested or representation this PR now emits.

resources/js/common/PMColumnFilterPopoverCommonMixin.js
CarliPinell: Observations were applied to fix the bug

resources/js/common/PMColumnFilterPopoverCommonMixin.js
CarliPinell: Observations were applied

resources/js/common/PMColumnFilterPopoverCommonMixin.js
CarliPinell: Observations were applied

resources/js/common/PMColumnFilterPopoverCommonMixin.js
CarliPinell: Observations were applied

resources/js/components/PMColumnFilterPopover/PMColumnFilterForm.vue
CarliPinell: Observations were applied

CarliPinell:
Filters are working as expected:

https://github.com/user-attachments/assets/37fcd784-731a-4828-b179-93a4ee26b04d

eiresendez (approved):
Approved since the main happy path from the ticket is working now: multiple `Status = ...` rows no longer behave as an `AND`, and the original repro is covered. 👍 

I’m attaching videos comparing `develop` vs this branch because I still see side effects with mixed status filters, for example `Status in [In Progress, Completed] OR Status = Self Service`. In that scenario, the branch appears to change the effective behavior and returns fewer rows than `develop`. This is worth checking before merge.

Develop:

https://github.com/user-attachments/assets/8a246e40-65d0-4699-845e-fe639a49933d

Branch bugfix/FOUR-26348:

https://github.com/user-attachments/assets/50f15683-5377-4b37-a79a-bddd6373a529

vladyrichter:
QA server K8S was successfully deployed https://ci-abbc08dd5a.engk8s.processmaker.net

PR #8798, opened 2026-04-27
bugfix/FOUR-30846: Completed tasks can be reassigned.

CarliPinell:
## Solution
- Reassign button is now disabled when task status is Completed or Closed.

## How to Test
- Login PM 4
- Run a Case from begin to end
- When process is completed, go to tasks
- Select a Completed process and open Preview Window
- Reassign button must be disabled.

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-30846

## Code Review Checklist

### DIFF
--- a/resources/js/tasks/components/TasksPreview.vue
+++ b/resources/js/tasks/components/TasksPreview.vue
@@ -166,13 +166,14 @@
                   <b-button
                     v-if="allowReassignment"
                     v-b-tooltip.hover
-                    class="btn text-secondary icon-button"
+                    class="btn text-secondary icon-button preview-reassign-btn"
                     variant="light"
                     :aria-label="$t('Reassign')"
                     :title="$t('Reassign')"
+                    :disabled="isTaskCompleted"
                     @click="openReassignment()"
                   >
-                    <i class="fas fa-user-friends" />
+                    <i class="fas fa-user-friends preview-reassign-btn__icon" />
                   </b-button>
                   <b-button
                     v-b-tooltip.hover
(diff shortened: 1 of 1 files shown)


resources/js/tasks/components/TasksPreview.vue
```diff
@@ -284,6 +285,26 @@ export default {
...
+    /**
+     * TasksList replaces status with HTML; API uses CLOSED for completed.
+     */
+    isTaskCompleted() {
+      const status = this.task?.status;
+      if (status === undefined || status === null) {
+        return false;
+      }
+      if (typeof status === "string") {
+        if (status === "CLOSED" || status === "Completed") {
```
eiresendez, line 298: 🟠 Improvements
resources/js/tasks/components/TasksPreview.vue:292
isTaskCompleted() mixes canonical API/DB state with UI display values: CLOSED, Completed, and rendered badge HTML/CSS classes.

Reasoning: CLOSED is the real task status. Completed is a UI label/alias and can be translated or absent depending on where the task object comes from. Parsing badge classes is also fragile.

Suggestion: if the frontend needs a visual disabled state, use a canonical field only, such as task.status === "CLOSED" or task.advanceStatus === "completed". Do not rely on translated labels or rendered HTML.

eiresendez (approved):
👍 👍 approved with just minor comments!

resources/js/tasks/components/TasksPreview.vue
CarliPinell: Sugestion is good. However, doing some tests on last CORE's develop branch, It seems "View eye" icon for Completed tasks was hidden from ellipsis menú so with this behavior this PR is not necessary.

PR #8803, opened 2026-04-28
FOUR-30819 Sanitize screen template config

eiresendez:
## Issue & Reproduction Steps
The default screen template "Single Form - Multi-Step Process" can persist serialized builder metadata in the target screen config. When the screen builder or preview renders that config, `mustache.render` can receive an object instead of a template string, causing a browser console error and preventing the screen from rendering correctly.

Reproduction:
1. Log in and go to Designer > Screens.
2. Create or open a screen.
3. Apply the default "Single Form - Multi-Step Process" screen template.
4. Save or reload the screen builder, then open preview/request mode.
5. Before this fix, the browser console reports `Invalid template! Template should be a "string" but "object" was given as the first argument for mustache#render`.

The regression test uses the existing `template-manifest-with-css-fields-layout.json` fixture, which contains serialized inspector metadata matching the failure mode.

## Solution
- Added backend sanitization for screen template config before applied templates are persisted.
- Removed serialized Vue component objects from `inspector[*].type` recursively.
- Normalized renderable string fields when template payloads provide non-renderable array/object values.
- Normalized empty `validation: []` values to `null` so form elements do not enter validation rendering paths with invalid template data.
- Kept the change backend-only; no public API, route, or frontend package changes were made.
- Added a regression test covering `POST /template/screen/{id}/apply` and nested sanitized config output.

## How to Test
Automated:
```bash
php ./vendor/bin/phpunit --filter testApplyTemplateSanitizesSerializedInspectorComponents tests/Feature/Templates/Api/ScreenTemplateTest.php
```

Manual:
1. Run `php artisan processmaker:sync-screen-templates` if the default screen templates need to be refreshed locally.
2. Create or open a screen in Designer.
3. Apply the default "Single Form - Multi-Step Process" screen template.
4. Save/reload the screen builder and open preview/request mode.
5. Confirm the browser console no longer shows the `mustache#render` invalid template error.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-30819

ci:deploy

### DIFF
--- a/ProcessMaker/Helpers/ScreenTemplateHelper.php
+++ b/ProcessMaker/Helpers/ScreenTemplateHelper.php
@@ -4,6 +4,32 @@ namespace ProcessMaker\Helpers;
 
 class ScreenTemplateHelper
 {
+    private const RENDERABLE_STRING_FIELDS = [
+        'ariaLabel',
+        'content',
+        'fieldValue',
+        'helper',
+        'label',
+        'loadingLabel',
+        'placeholder',
+    ];
+
+    /**
+     * Remove serialized Vue component definitions from screen config.
+     *
+     * Screen templates can contain old inspector metadata where inspector.type
+     * is a serialized Vue component object. That data is not needed at runtime
+     * and can reach renderer paths that expect Mustache templates to be strings.
+     */
+    public static function sanitizeScreenConfig($config): array
+    {
+        if (!is_array($config)) {
+            return [];
+        }
+
+        return self::sanitizeConfigValue($config);
+    }
+
     /**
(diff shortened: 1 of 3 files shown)


Kookster310:
QA server K8S was successfully deployed https://ci-9a779e9720.engk8s.processmaker.net

Kookster310:
QA server K8S was successfully deployed https://ci-9a779e9720.engk8s.processmaker.net

mcraeteisha (approved):

Kookster310:
QA server K8S was successfully deployed https://ci-9a779e9720.engk8s.processmaker.net

PR #8825, opened 2026-05-11
FOUR-31158 Resolve group assignment imports by name

eiresendez:
## Issue & Reproduction Steps
FOUR-31158 fixes group task assignment resolution during v2 process import/export.

Reproduction:
1. On server A, create a process with a task assigned to a group.
2. Export the process.
3. On server B, create or keep groups where UUIDs and names do not fully match the exported source group.
4. Import the process on server B.

Before this fix, discarded group assignment references could resolve by UUID even when the target group name was different, which could assign imported tasks to the wrong group. The fixture pattern used by the automated tests is `process-with-different-kinds-of-assignments.bpmn.xml` with group assignment dependents.

## Solution
- Added an optional discarded dependent resolver hook in `Dependent`.
- Added a `GroupExporter` resolver for discarded `group_assignment` dependents that resolves by exact group name only.
- Preserved existing behavior for normal full asset imports and for non-group-assignment discarded references.
- Added coverage for UUID/name match, name-only match, UUID-only mismatch, missing groups, and mixed multiple group assignments.

## How to Test
Run:

```bash
./vendor/bin/phpunit tests/Feature/ImportExport/Exporters/AssignmentExporterTest.php
```

Manual validation:
1. Export a process with group task assignments from server A.
2. Import it into server B.
3. Confirm tasks assign to existing target groups when the group name matches.
4. Confirm tasks remain unassigned when only the UUID matches but the group name differs, or when no matching group name exists.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-31158
- https://processmaker.atlassian.net/browse/FOUR-29998

ci:deploy

### DIFF
--- a/ProcessMaker/ImportExport/Dependent.php
+++ b/ProcessMaker/ImportExport/Dependent.php
@@ -8,7 +8,7 @@ class Dependent
         public string $type,
         public string $uuid,
         public Manifest $manifest,
-        public $meta,
+        public mixed $meta,
         public string $exporterClass,
         public string $modelClass,
         public array $fallbackMatches,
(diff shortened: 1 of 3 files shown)


eiresendez:
Reviewer testing recommendation: I recommend validating this with the multitenant feature so you can see the before/after behavior clearly. Export a process with group task assignments from server A, import it into server B, and confirm that when groups already exist on server B with the same name, task assignments resolve by group name instead of UUID.

Kookster310:
QA server K8S was successfully deployed https://ci-41605cfc76.engk8s.processmaker.net

nolanpro (approved):

PR #8835, opened 2026-05-18
FOUR-31336 | Change on clients adding modal for plainSecret

AugustoLopezProcess:
## Issue & Reproduction Steps
Upgrading Laravel Passport stores OAuth client secrets as **hashed** values only. The plain-text secret is available once at creation via `$client->plainSecret` and is not persisted in a readable form.
**Before this change:**
1. Go to **Admin → Auth Clients** (`/admin/auth-clients`).
2. Create a new auth client.
3. The listing showed a **Client Secret** column (truncated secret + copy button).
4. After the Passport upgrade, listing/show responses no longer expose a usable plain-text secret, and the old UI implied secrets could be viewed or copied later.
**Expected:** Users must see and save the plain-text secret **only at creation time**, consistent with Passport’s one-time `plainSecret` behavior.

---

## Solution
- **`AuthClient` API resource** — Return `secret` from `$this->plainSecret` instead of `$this->secret`, so create responses include the one-time plain-text value while list/show responses do not expose a readable secret.
- **`ClientController`** — Remove `$client->makeVisible('secret')` after store; it is no longer needed with hashed secrets.
- **Auth Clients listing (`AuthClientsListing.vue`)** — Remove the **Client Secret** column, its copy-to-clipboard slot, and related styles/helpers.
- **Auth Clients page (`index.blade.php`)** — After a successful create/save, open a **Secret** modal with:
  - Read-only secret field populated from `response.data.secret`
  - Copy-to-clipboard button
  - Warning: *"This is the only time you will be able to view the client secret. Keep it in a safe place."*
- **i18n** — Add English/Spanish strings for the warning, modal copy action, and related labels.

## How to Test
1. Log in as an admin user with access to **Auth Clients**.
2. Open `/admin/auth-clients`.
3. Confirm the table has **Client ID**, **Name**, **Redirect**, and actions — **no Client Secret column**.
4. Click **Auth Client** and create a client (e.g. Authorization Code Grant with a valid redirect URL).
5. After save:
   - The create modal closes.
   - A **Secret** modal appears with the full plain-text secret.
   - The amber warning message is visible.
   - **Copy** puts the secret on the clipboard.
6. Close the secret modal; the listing refreshes and does **not** show the secret.
7. Edit an existing client and save — confirm behavior matches expectations (secret modal only when the API returns a `secret`; after creation, list/edit should not reveal a readable secret).
8. Optionally verify API behavior:
   - `POST /oauth/clients` → response includes `secret` (plain text) on create.
   - `GET /oauth/clients` → items do not include a usable plain-text `secret`.

---

## Related Tickets & Packages
- Related to Laravel Passport upgrade (hashed client secrets / `plainSecret`).
- [FOUR-31336](https://processmaker.atlassian.net/browse/FOUR-31336?focusedCommentId=509754)

---

## Code Review Checklist

[FOUR-31336]: https://processmaker.atlassian.net/browse/FOUR-31336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nolanpro (changes requested):
@AugustoLopezProcess when I edit and save an existing client, it still shows the warning but without the secret. It shouldn't show the popup when editing an existing client.

Image

AugustoLopezProcess (commented):
Preventing now that the modal shows on edit, and refresh listing of clients if edit

nolanpro (approved):

PR #8842, opened 2026-05-26
FOUR-31336 Add swagger-ui asset route

nolanpro:
Looks like swagger-ui made an update that's not compatible with l5-swagger. This is a workaround. See https://github.com/DarkaOnLine/L5-Swagger/issues/648

https://processmaker.atlassian.net/browse/FOUR-31336


gusys (approved):
Looks good to me

PR #8846, opened 2026-05-29
FOUR-31631 Fix security advisories

nolanpro:
https://processmaker.atlassian.net/browse/FOUR-31631

- Removes 2 packages from composer that are not explicitly used by our code or our packages. They are used by other dependencies so let them handle that.
- Ran composer update on a number of vulnerabile symfony components that are dependencies of laravel/framework (but the specific fixed versions were not specified yet in laravel).


AugustoLopezProcess (approved):
Looks good to me

PR #8895, opened 2026-07-06
FOUR-31748: Fix Investigate increased IOPS reads on EFS

caleeli:
## Solution
- Reduce the severity of log when executing a system process like ABE


julceslauhub (approved):
Approved

PR #8896, opened 2026-07-07
FOUR-26363 | +Case Pagination Is Not Working Correctly

mcraeteisha:
# Issue
Ticket: [FOUR-26363](https://processmaker.atlassian.net/browse/FOUR-26363)

This PR fixes the +Case modal per-page selector so that switching between 15 / 30 / 50 immediately refreshes the visible process list.

# Solution
`requestModal.vue` inherits `changePerPage()` from `datatableMixin`, which reads `this.data.meta.total` to clamp the current page. But this component never assigns `this.data`; `fetch()` only writes the response into `this.$refs.listProcess.data` and `this.processes`. So `this.data.meta` was `undefined`, the shared handler threw before reaching `this.fetch()`, and the list never refreshed.

The solution was to override `changePerPage(value)` in `resources/js/components/requests/requestModal.vue` to read the total from `this.$refs.listProcess.tablePagination` (the same source `onPageChange` already uses), guard against the modal's lazily-mounted ref, clamp `page`, then call `fetch()`. Mirrors the pattern already used in `CasesRetentionLogs.vue`, with extra null-safety because this component lives inside a `<b-modal>` and doesn't fetch on `created`.

# How To Test
1. Login to ProcessMaker. Open +Case modal → verify list shows the default 15 per page.
2. Change per-page selector to 30.
	   - Ensure the list refreshes immediately, pagination shows "1 – 30 of N" (if 30+ processes exist).
3. Change per-page selector to 50.
     - Ensure the list refreshes immediately and pagination reflects 50 (if 50+ processes exist).
4. Change back to 15.
     - Esnure the list refreshes and pagination reflects 15.
5. After each per-page change, use `<`, `>`, `<<`, `>>`, and direct page numbers to ensure paging still works.
6. Type in the search box and ensure the filter still resets to page 1 and fetches correctly.
7. Close and reopen the modal. Ensure the initial fetch works, per-page defaults to 15, all of the above still works.

ci:deploy

# Code Review Checklist
- [ ]  This code adheres to ProcessMaker Coding Guidelines.

[FOUR-26363]: https://processmaker.atlassian.net/browse/FOUR-26363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


resources/js/components/requests/requestModal.vue
```diff
@@ -192,6 +192,18 @@ export default {
       }
       this.fetch();
     },
+    changePerPage(value) {
```
eiresendez, line 207: The new changePerPage() method has an off-by-one page clamp when the total is an exact multiple of the selected page size. The current code uses `Math.floor(total / value) + 1`, so with `total = 30` and `value = 30` it produces an impossible footer state: `31 - 60 of 30 Processes`. This should use the actual last page instead, for example `Math.ceil(total / Number(value))`.

resources/js/components/requests/requestModal.vue
```diff
@@ -192,6 +192,18 @@ export default {
...
+      this.perPage = value;
```
eiresendez: Closing and reopening the modal should return to the initial state with the default `15` per page, but `showModal()` currently does not reset page or perPage; it only sets loaded, shows the modal, and calls `fetch()`. If a user changes the selector to 30 or 50, closes the modal, and opens it again, the component can keep the previous pagination state instead of starting from the expected default.

eiresendez (changes requested):

eiresendez (approved):
👍 👍  Working as expected!

vladyrichter:
QA server K8S was successfully deployed https://ci-91e619443e.engk8s.processmaker.net

mcraeteisha:
Please note that the security check failure by GitGuardian comes from `reset.blade.php` which is unrelated to the code changes in this PR.

PR #8918, opened 2026-07-20
FOUR-32305: Export assets referenced in scripts using environment variables

caleeli:
## Issue & Reproduction Steps
As process architect I require us export/import assets referenced in script code same as we do in other places.

## Solution
- implement asset linking for environment variables

## How to Test
1. Create a collection
2. Create an environment variable
3. Link this environment variable with the created collection
4. Use this environment variable in a script
5. Use the script in a process
6. Export the process
7. Import the process in a different place
8. Check that the collection ID corresponds to the right one

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-32305

ci:deploy


vladyrichter:
QA server K8S was successfully deployed https://ci-8568b9d751.engk8s.processmaker.net

vladyrichter:
QA server K8S was successfully deployed https://ci-8568b9d751.engk8s.processmaker.net

caleeli:
Source Server:
image

caleeli:
Exporting the process with linked environment variable:


https://github.com/user-attachments/assets/2433f09a-f4e5-43a9-8e70-935d986d4c2f



Importing the process with linked environment variable:

https://github.com/user-attachments/assets/0cee37ba-aa88-49b6-8a97-883705c0a987

ProcessMaker/ImportExport/Exporters/EnvironmentVariableExporter.php
```diff
@@ -17,10 +19,51 @@ class EnvironmentVariableExporter extends ExporterBase
...
+        }
+
+        $this->addDependent(DependentType::ENVIRONMENT_VARIABLE_ASSET, $asset, $exporterClass);
     }
 
     public function import() : bool
     {
+        foreach ($this->getDependents(DependentType::ENVIRONMENT_VARIABLE_ASSET, true) as $dependent) {
+            $asset = $dependent->model;
+            if (!$asset) {
```
eiresendez, line 55: 🔴 A discarded or unresolved asset (Custom Import) can be returned as an empty model. The truthiness check at line 55 passes, then the importer saves `asset_uuid = null` and an empty numeric value. The fallback at line 66 is therefore bypassed, leaving scripts with a broken environment variable after Import or DevLink. Only use a persisted dependent asset; otherwise fail clearly or restore a deliberately unlinked value without persisting a half-link.

resources/js/processes/environment-variables/assetTypes.js
```diff
@@ -0,0 +1,54 @@
...
+  {
+    class: "ProcessMaker\\Packages\\Connectors\\DataSources\\Models\\DataSource",
+    label: "Data Connector",
+    apiPath: "data_sources",
+    nameField: "name",
+  },
+  {
+    class: "ProcessMaker\\Package\\PackageDecisionEngine\\Models\\DecisionTable",
+    label: "Decision Table",
+    apiPath: "decision-tables",
```
eiresendez: 🔴 This point to nonexistent APIs. Decision Tables use `decision_tables`; FlowGenies use `package-ai/flow_genies`. Because [AssetLinkFields.vue (line 192)](/Users/eresendez/projects/processmaker/processmaker/resources/js/processes/environment-variables/components/AssetLinkFields.vue:192) silently converts request failures into an empty list, both advertised asset types are unusable without any visible error.

database/migrations/2026_07_20_000000_add_asset_link_to_environment_variables_table.php
```diff
@@ -0,0 +1,29 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('environment_variables', function (Blueprint $table) {
+            $table->string('asset_type')->nullable()->after('value');
+            $table->uuid('asset_uuid')->nullable()->after('asset_type');
```
eiresendez: 🟠 `asset_uuid` is not required by the import/export framework as currently designed. FOUR-32305 specifies `asset_type` plus the existing `value`, which already holds the local asset ID. The exporter can resolve that model using those two fields, and `addDependent()` obtains its UUID at runtime for the manifest. Import then writes the target model’s numeric ID into `value`. Unless David identifies a separate product requirement for a persisted stable relationship, I recommend removing `asset_uuid`; otherwise the PR should document which field is authoritative and why both must exist.

ProcessMaker/Http/Controllers/Api/EnvironmentVariablesController.php
```diff
@@ -172,14 +174,21 @@ public function show(EnvironmentVariable $environment_variable)
      */
     public function update(EnvironmentVariable $environment_variable, Request $request)
     {
-        $fields = ['name', 'description'];
-        if ($request->filled('value')) {
+        $data = $this->prepareAssetLinkInput($request->all());
```
eiresendez, line 180: 🟠 If `asset_uuid` is retained, update validation does not actually guarantee both-or-neither. It validates only the incoming payload, so clearing one link field while omitting the other can persist an inconsistent pair. Validate the merged resulting state or update both fields atomically.

eiresendez (changes requested):
Hi @caleeli, I left a few comments that I think are worth considering. If they make sense, let me know what you think.
cc: @nolanpro

database/migrations/2026_07_20_000000_add_asset_link_to_environment_variables_table.php
caleeli: all the exportable assets include its uuid to identify if they are the same in any other server, storing uuid also the table follow the pattern from other assets like collections, screens, processes, so I thought it is better to keep the reference that the value column

ProcessMaker/Http/Controllers/Api/EnvironmentVariablesController.php
caleeli: Added the validation of assets linking:

Image

If the asset does not match a valid one returns an 422

vladyrichter:
QA server K8S was successfully deployed https://ci-8568b9d751.engk8s.processmaker.net

database/migrations/2026_07_20_000000_add_asset_link_to_environment_variables_table.php
caleeli: Thanks @eiresendez the asset_uuid is resolved by Exporter Base so it is not needed. This PR has the changes with using asset_uuid: https://github.com/ProcessMaker/processmaker/pull/8928

ProcessMaker/ImportExport/Exporters/EnvironmentVariableExporter.php
caleeli: This scenario is covered and tested in this PR: https://github.com/ProcessMaker/processmaker/pull/8928

resources/js/processes/environment-variables/assetTypes.js
caleeli: Paths fixed. Also fixed in PR: https://github.com/ProcessMaker/processmaker/pull/8928

vladyrichter:
QA server K8S was successfully deployed https://ci-8568b9d751.engk8s.processmaker.net

vladyrichter:
QA server K8S was successfully deployed https://ci-8568b9d751.engk8s.processmaker.net

nolanpro:
Closed in favor of https://github.com/ProcessMaker/processmaker/pull/8928

PR #8920, opened 2026-07-21
FOUR-32304: Prevent duplicate DevLinks from breaking bundle lists

eiresendez:
## Issue & Reproduction Steps
DevLink allows multiple Linked Instance records with different names to point to the same destination URL. Connecting the newer duplicate replaces the OAuth client used by the previous connection, leaving installed bundles associated with stale credentials. Their remote version checks can then return errors that surface as local HTTP 500 responses and prevent the Local Bundles list from loading.

To reproduce, link a remote instance and install bundles from it, add the same destination URL again under a different name, and complete the new connection. Open Admin > DevLink > Local Bundles and observe the failed version checks for bundles associated with the stale connection. The QA recordings are attached to FOUR-32304. Manual local fixtures named `FOUR-32304 REPRO - Legacy Bundle 1` and `FOUR-32304 REPRO - Legacy Bundle 2`, linked to the stale DevLink with remote IDs 1 and 2, were also used to validate the legacy-data behavior.

## Solution
- Make Linked Instance creation create-only, normalize HTTP/HTTPS origins, and reject duplicate names or normalized URLs with an inline `422` validation message.
- Move creation validation into a dedicated FormRequest and HTTP-origin validation rule.
- Return a controlled `200` response with `available: false` and `version: null` when remote version checks fail, while marking successful responses as available.
- Keep legacy bundles visible, display `Connection unavailable`, and disable Update, Reinstall, and Add a Copy until their associated connection is available.
- Preserve existing duplicate records without a migration, unique database index, or automatic data cleanup.

## How to Test
1. Run `vendor/bin/phpunit tests/Feature/Admin/DevLinkTest.php tests/Model/DevLinkTest.php tests/Feature/Api/DevLinkTest.php --no-coverage`.
2. Run `npm run development`.
3. Create a Linked Instance using a new URL and confirm the stored URL is normalized.
4. Attempt to create another Linked Instance using the same URL with case, trailing-slash, or default-port variations. Confirm the modal displays the duplicate warning and does not start OAuth.
5. Load Local Bundles with legacy duplicate DevLinks and revoked credentials. Confirm the complete list remains visible, affected bundles show `Connection unavailable`, and remote actions are disabled.
6. Reconnect the DevLink associated with an installed bundle. Confirm existing remote bundles can show `Update Available`, missing remote bundles remain unavailable, and remote actions are enabled only for available bundles.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-32304
- https://processmaker.atlassian.net/browse/FOUR-32196


ProcessMaker/Rules/HttpOrigin.php
```diff
@@ -0,0 +1,29 @@
...
+
+            return;
+        }
+
+        $path = $parts['path'] ?? '';
+        if (
+            !isset($parts['scheme'], $parts['host'])
+            || !in_array(strtolower($parts['scheme']), ['http', 'https'], true)
+            || !in_array($path, ['', '/'], true)
+            || isset($parts['query'], $parts['fragment'], $parts['user'], $parts['pass'])
```
CarliPinell: `isset($parts['query'], $parts['fragment'], $parts['user'], $parts['pass'])` is true only when **all** of those keys exist. A URL with only a query (`https://host.test/?x=1`), only a fragment, or only credentials (`https://user:pass@host.test`) will incorrectly pass this rule.

Because `normalizeUrl()` later rebuilds from scheme/host/port only, those values are silently stripped instead of rejected.

Suggest checking each key independently, e.g.:

```php
|| isset($parts['query'])
|| isset($parts['fragment'])
|| isset($parts['user'])
|| isset($parts['pass'])
```

CarliPinell (commented):
Reviewed FOUR-32304. One validation issue worth fixing before merge; otherwise the DevLink create-only + remote version availability approach looks solid.

CarliPinell (changes requested):

ProcessMaker/Rules/HttpOrigin.php
eiresendez: Thanks @CarliPinell  issue resolved.

CarliPinell (approved):
Approved.

PR #8927, opened 2026-07-22
FOUR-32333: Isolate DevLink progress by operation

eiresendez:
## Issue & Reproduction Steps
DevLink installation and update progress is broadcast through a shared private user channel. When the same user starts operations from different tabs or browsers, each open progress modal currently consumes every event and can overwrite its original progress or result with the status of another operation.

To reproduce:
1. Sign in with the same user in two browser tabs or browsers.
2. In the first tab, start a Bundle installation or update and leave its progress modal open.
3. In the second tab, start another Bundle or asset operation.
4. Return to the first tab and observe that its progress and result can be replaced by events from the second operation.

The Jira ticket includes a video demonstrating the issue; no additional fixture is required.

## Solution
- Added a unique operation identifier to Bundle installation, update, reinstall, and asset installation requests.
- Propagated the identifier through `DevLinkInstall`, the import/export logger, and `ImportLog` events.
- Filtered progress events in each modal so it only processes events belonging to its own operation.
- Remounted the progress component for every attempt to prevent inherited progress, warnings, or errors.
- Preserved compatibility by generating an identifier when older API consumers omit it and keeping the logger and queued job identifier optional.
- Added API, logger, and failed-job regression coverage for operation identifier propagation.

## How to Test
1. Sign in as the same user in two tabs or browsers.
2. Start a Bundle installation or update in the first tab.
3. Start another Bundle reinstall or asset installation in the second tab.
4. Verify that each modal only displays its own progress, messages, and final result.
5. Verify that starting a new attempt opens a clean progress modal without state inherited from the previous attempt.
6. Call an installation endpoint without `operation_id` and verify that the existing `{"status":"queued"}` response remains unchanged and the backend generates an identifier.

Verification completed:
- Focused PHPUnit suites passed: 11 tests and 26 assertions.
- `npm run development` compiled successfully.
- Manual same-user, two-tab testing confirmed that while operation A was at 83%, operation B remained at 0% until its own job began; both then completed independently.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-32333

ci:deploy


vladyrichter:
QA server K8S was successfully deployed https://ci-42e4078cfe.engk8s.processmaker.net

vladyrichter:
QA server K8S was successfully deployed https://ci-42e4078cfe.engk8s.processmaker.net

mcraeteisha (approved):

PR #8941, opened 2026-07-27
Update app.instance configuration for tenant

nolanpro:
The app instance is used for the docker image name for custom executors. We have always set it to the database name so let's do the same thing for tenant executors


gusys (approved):
Looks good to me...

PR #8945, opened 2026-07-28
FOUR-32421 Fix for init microservice and transition command

nolanpro:
- Skip php-nayra
- Don't mark executors a duplicate
- Don't mark executors as system


gusys (approved):
Looks good to me...

PR #8984, opened 2026-08-10
FOUR-32618 Command TENANT=<tenant-id> php artisan route:cache is not working

julceslauhub:
https://processmaker.atlassian.net/browse/FOUR-32618


nolanpro (changes requested):
@julceslauhub I get that we can't use https://github.com/spatie/laravel-multitenancy/blob/main/src/Tasks/SwitchRouteCacheTask.php because we're saving it in tenant-specific storage, but could you integrate this part so it works in octane?
```
if (app()->routesAreCached() && $this->shouldReinitializeRouter()) {
            require app()->getCachedRoutesPath();
        }
```

julceslauhub:
Solved with this another PR https://github.com/ProcessMaker/processmaker/pull/8993

PR #9009, opened 2026-08-25
fix(FOUR-32802): prevent Available Columns timeout under Octane

gproly:
Skip getDataColumnsAttribute() when VariableFinder is enabled and merge process_columns with paginated variables instead. Fix empty processIds query filter and avoid double pagination

https://processmaker.atlassian.net/browse/FOUR-32802


rodriquelca:
Hello team 
@gproly please add the test unit to ensure this solution and improve the test coverage
thanks

PR #9022, opened 2026-08-31
FOUR-31145: Self Service tasks not visible to users in assigned subgroups

rodriquelca:
## Issue & Reproduction Steps
A Self Service task assigned to group Main does not show for a user who is only in subgroup Sub (Sub ∈ Main). Inbox/claim compared the token’s self_service_groups to the user’s direct groups only.

The task list also broke: exclude() used $this->table (null) and generated SELECT ``.id``, so the API returned 422 and the tab looked empty.

Repro: Homero in Sub → Sub in Main → process Self Service assigned to Main → login as Homero → Tasks → Self Service. List is empty; claim fails. To Do empty until claim is expected.

## Solution

- User::selfServiceGroupIds() = direct groups + ancestors (Group::ancestorIdsFor()). Used by availableSelfServiceTasksQuery() and canSelfServe(). Token is not rewritten.
- Cache 10 min; GroupMemberObserver invalidates on membership changes.
- scopeExclude() uses getTable() so the list SELECT is valid.

https://github.com/user-attachments/assets/08ed8e58-3ada-499b-933e-724caf198730

## How to Test
Run the  unit test
./vendor/bin/phpunit tests/Feature/SelfServiceSubgroupTest.php

## Related Tickets & Packages
https://processmaker.atlassian.net/browse/FOUR-31145

## Code Review Checklist

ci:deploy

### DIFF
--- a/ProcessMaker/Models/Group.php
+++ b/ProcessMaker/Models/Group.php
@@ -2,6 +2,8 @@
 
 namespace ProcessMaker\Models;
 
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Cache;
 use Illuminate\Validation\Rule;
 use ProcessMaker\Models\EmptyModel;
 use ProcessMaker\Query\Traits\PMQL;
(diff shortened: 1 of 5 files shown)


nolanpro:
QA server K8S was successfully deployed https://ci-56d7fe3332.engk8s.processmaker.net

pmPaulis (approved):

ProcessMaker/Models/Group.php
```diff
@@ -156,4 +158,63 @@ public function assigned()
...
+
+        while ($queue) {
+            $parents = GroupMember::query()
+                ->where('member_type', self::class)
+                ->whereIn('member_id', $queue)
+                ->pluck('group_id')
+                ->all();
+
+            $queue = [];
+            foreach ($parents as $parentId) {
```
pmPaulis, line 207: @rodriquelca Could changing this foreach to a while loop cause any performance issues? Could you analyze it and also check what happens if a group has a recursive assignment?

caleeli: @rodriquelca consider to add a test with circular reference of groups:
```
Group0 -> Group1 -> Group0
```

Seems the code covers it but please include it in a test

caleeli (changes requested):
Add a test according to the previous comment (circular references)

ProcessMaker/Models/Group.php
rodriquelca: @pmPaulis 
We could use a while in the inner loop too — it would behave the same. We kept foreach because it’s the normal PHP way to loop over a list of results.

The important part for performance is the outer while: it does one DB query per hierarchy level. The inner loop only reads that result in memory, so foreach vs while doesn’t matter here.

Circular groups are safe: $visited skips groups we’ve already seen, so we never loop forever.

rodriquelca: @caleeli  circular test was added

vladyrichter:
QA server K8S was successfully deployed https://ci-56d7fe3332.engk8s.processmaker.net

PR #9034, opened 2026-09-03
feat(FOUR-32734): allow ProcessMaker login in cross-site iframes

gproly:
Set CSP frame-ancestors and relax session cookies (SameSite=None, Secure, partitioned) when IFRAME_EMBEDDING_ENABLED=true and IFRAME_EMBEDDING_ALLOWED_ORIGINS lists parent origins (e.g. http://localhost:8093).

Example: pmList embeds PM login via
<iframe src="https://127.0.0.5:8092/login" ...></iframe>.

Security notes:
- Only enable when cross-site embedding is required; disable when not in use.
- IFRAME_EMBEDDING_ALLOWED_ORIGINS must list explicit parent origins; empty allowlist does not activate relaxed cookies or CSP.
- SameSite=None increases cross-site cookie exposure; CSRF risk is higher—restrict allowed origins to trusted domains only.
- In production, only https:// origins are accepted for frame-ancestors (http:// entries are ignored).
- Third-party cookie blocking in browsers may still break iframe login; test target browsers and prefer HTTPS on both parent and PM.
- Do not use as a substitute for SSO; treat as a narrow integration option.

https://processmaker.atlassian.net/browse/FOUR-32734

### DIFF
--- a/ProcessMaker/Http/Middleware/HideServerHeaders.php
+++ b/ProcessMaker/Http/Middleware/HideServerHeaders.php
@@ -72,7 +72,42 @@ class HideServerHeaders
             $response->headers->set('Server', 'ProcessMaker Server');
         }
 
+        $this->allowIframeEmbedding($response);
+
         return $response;
+    }
+
+    private function allowIframeEmbedding(Response $response): void
+    {
+        $origins = trim((string) env('IFRAME_EMBEDDING_ALLOWED_ORIGINS', ''));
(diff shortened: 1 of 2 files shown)


pmPaulis:
@gproly 
All the changes related to this configuration needs to add in the package-product-analytics.
In sync wiht support we will add the considerations to know for this change
Regards,

